Issue #9, opened 2019-06-21
Connect Graph API


cdcabrera:
## Feature description

As a developer I'd like to view actual data associated with the dashboard graph.

### Assumptions and Questions
- API response schema should match proposed mocks
  - confirm authentication header formats for interacting with the API
  - confirm parameters
  - confirm response properties

## Acceptance Criteria
- [x] Initial base redux layout, tests against mock data laid out
- [x] Verify that the appropriate headers are being set
  - authorization/tokens
  - environment, if required
- [x] Verify query parameters are available, and can be accessed through Redux
- [x] Verify that responses are being consumed through the Redux state management layer
  - If available, also verify that the responses match proposed examples within proposed OpenApi spec
- [x] Verify service and Redux unit tests have been updated

### Phase 1 work
* Phase 1 should include connecting to the RHSM API. This story assumes a spec/examples/mocks being minimally available for display consumption
#### Cloud Meter
Cloud Meter's api will not be consumed during phase 1
* Reference [cloudigrade for report params and commands](https://gitlab.com/cloudigrade/cloudigrade/blob/master/docs/rest-api-examples.rst)
* Reference [cloudigrade, accessing the APIs](https://gitlab.com/cloudigrade/cloudigrade/wikis/how-do-i-reach-these-apis)

PR #37, opened 2019-07-16
fix(build, rhelServices, auth): issues/9 build, rhsm api updates 

cdcabrera:
## What's included

- fix(build, rhelServices, auth): issues/9 build, rhsm api updates 
  * build, dotenv api paths breakout, restructure
  * build, spandx config, staging, app, and api paths
  * build, dist test double check for localhost refs
  * build, npm script proxy run, remove mock server
  * rhelApiTypes, rhelGraphReducer, rhelActions, rhsm updates
  * rhelServices, remove cloud meter refs, update rhsm mock and func
  * authentication, fix for insights global and remove listeners
  * helpers, display name

### Notes

- We updated the RHSM API path for the proxy, however the response ends up timing out. This may be due to not passing the headers, which in turn relates to how we're not integrated into the navigation under the Insights Chrome. 
- We did include some basic updates to the Authentication component that helps fix the login display, once we are integrated into the Insights Chrome nav.

1. `$ yarn test`

### Run the mock server with RHSM
1. `$ yarn start`
1. open the browser debugging console, network tab and look for `RHEL`, mock data should be accessible
### Run the proxy server with RHSM
1. `$ yarn start:proxy`
1. open the browser debugging console, network tab and look for `RHEL`, data through the API should be accessible
### Run the build
1. `$ yarn build`
1. and all the tests should pass...

## Example

The mock response from the local run `$ yarn start` should look something like
```
{
  "data": [
     {
       "cores": 56,
       "date": "2019-06-01T00:00:00Z",
       "instance_count": 28
     },
     {
       "cores": 52,
       "date": "2019-06-02T00:00:00Z",
       "instance_count": 26
     },
     ...
   ],
   "links": {
     "first": null,
     "last": null,
     "previous": null,
     "next": null
   },
   "meta": {
     "resultSetSize": 30,
     "product": "RHEL",
     "granularity": "DAILY"
   }
}
```

## Updates issue/story

Updates #9


priley86:
this looks fantastic to me. Cloned and ran locally, tests pass and UI runs fine.

One minor bit we can probably improve on in future increasing coverage in `authenication.js` some way. Maybe abstracting those insights API calls to another service would make it easier to test? Not a big deal right now and should not hold up other work!

🚂  🚂 🚂

cdcabrera:
Oddly we did give a flavor of that a go, some of it ended up being a similar thing just a different file... Howweeeeeever, I didn't really see it at the time, but now I can see us moving the insights aspects alongside aspects alongside, or within, the `authorizeUser` function inside of `userServices`... that would wipe out a couple of more references, hmmmmmmm, good call @priley86 !

PR #55, opened 2019-07-24
feat(locale): use cookie to determine current locale

### LINKED ISSUE #44: Get locale from JWT or locale cookie
## Description
As a developer it would be more helpful to an end user to not hardcode the loaded locale. Instead it should be loaded either from the available JWT or the locale cookie.

## Acceptance Criteria
- [ ] locale preferences should load from the available JWT or locale cookie
- [ ] this update should be incorporated into `userServices.js` and make use of the `js-cookie` npm package

## Assumptions and Questions
- the JWT and/or locale cookie can be found by using the proxy interface, logging in, and using your browsers developer console around some flavor of storage settings

priley86:
## What's included

    * use rh_locale cookie to set i18n locale if it's available
    * use env variables as a fallback if we can't find a supported locale

Valid i18n codes (and major codes):
https://github.com/ladjs/i18n-locales#tip

## How to test

2. `$ yarn test:dev`

## Example

.

## Updates issue/story

closes #44


src/services/__tests__/userServices.test.js
```diff
@@ -19,4 +20,27 @@ describe('UserServices', () => {
...
   });
+
+  it('should return default locale if no locale cookie is present', done => {
+    userServices.getLocale().then(locale => {
+      expect(locale).toMatchSnapshot();
+      done();
+    });
+  });
+
+  it('should return en locale if cookie is set to en_US', done => {
```
cdcabrera: optional: maybe make a more generic description... `should return a specific locale cookie value`

src/services/__tests__/userServices.test.js
```diff
@@ -19,4 +20,27 @@ describe('UserServices', () => {
...
+    Cookies.get = jest.fn().mockImplementation(() => 'en_US');
+    userServices.getLocale().then(locale => {
+      expect(locale).toMatchSnapshot();
+      done();
+    });
+  });
+
+  it('should return test locale if cookie is set to test_US', done => {
```
cdcabrera: request: lets change the description on this one tooooo: `should return the default locale with an invalid ISO_639 code`

.env
```diff
@@ -5,8 +5,10 @@ PUBLIC_URL=${UI_DEPLOY_PATH_PREFIX}/apps/subscription-reporting/
 
 REACT_APP_AJAX_TIMEOUT=60000
 
+REACT_APP_CONFIG_SERVICE_LOCALES_COOKIE=rh_locale
 REACT_APP_CONFIG_SERVICE_LOCALES_DEFAULT_LNG=en
-REACT_APP_CONFIG_SERVICE_LOCALES_DEFAULT_LNG_DESC=English
+REACT_APP_CONFIG_SERVICE_LOCALES_DEFAULT_LNG_KEY=en_US
```
cdcabrera: request: we'll flip this back to the `...DESC=English` based on discussions

.env
```diff
@@ -5,8 +5,10 @@ PUBLIC_URL=${UI_DEPLOY_PATH_PREFIX}/apps/subscription-reporting/
 
 REACT_APP_AJAX_TIMEOUT=60000
 
+REACT_APP_CONFIG_SERVICE_LOCALES_COOKIE=rh_locale
 REACT_APP_CONFIG_SERVICE_LOCALES_DEFAULT_LNG=en
```
cdcabrera: request: 
- this'll change to `en-US` along with a couple of other locations...
- `./public/locales/en.json` --> en-US.json
- `./public/locales/locales.json` --> the English references inside need updates

research:
we may be able to have the i18n plugin strip the locale subset for us with a config setting... meaning we wouldn't have to update anything going from `en` -> `en-US`
- https://www.i18next.com/
- https://react.i18next.com/

package.json
```diff
@@ -84,6 +84,7 @@
     "axios": "^0.19.0",
     "i18next": "^17.0.6",
     "i18next-xhr-backend": "^3.0.0",
+    "js-cookie": "^2.2.0",
```
cdcabrera: request: npm `locale-code`

priley86: done

src/services/userServices.js
```diff
@@ -14,19 +15,37 @@ const authorizeUser = () => {
...
+      return { value: majorLocaleValue, key: cookieValue };
+    }
+    return null;
+  } catch (e) {
+    if (!helpers.TEST_MODE) {
+      console.warn(`Unable to parse locale ${e.message}`);
+    }
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
priley86: "future proof"... would love if the platform offered the set of supported lng codes eventually like you mention...

src/services/userServices.js
```diff
@@ -14,19 +15,37 @@ const authorizeUser = () => {
...
     if (locale) {
       return resolve({
-        data: locale
+        data: getLocaleFromCookie() || defaultLocale
       });
     }
-
     return resolve({});
   });
 };
```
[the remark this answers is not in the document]
priley86: so this gets weird... i've gone ahead and updated to use `en-US.json` b/c that's what we discussed, and I think it's the most readable...and will totally work....

BUTTTT... i18nnext will by default translate these four/five letter codes into "fallback" two letter codes... so it will request both `en.json` and `en-US.json` as the primary once it does not find `en.json`... saw that logic in the debugger here after testing this:
https://github.com/i18next/i18next/blob/270904f6369ee9bbda059c3186fcea7baf9eb15d/src/LanguageUtils.js#L33

All this being said... i still prefer to "key" on the four/five letter codes. Better specificity for future possibilities of supporting those minor locales. Also... having a "major" locale fallback file could also be a thing and just supported out of the box w/ this kind of configuration. Pretty cool it does this!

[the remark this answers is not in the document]
cdcabrera: Have 2 easy solutions 
Either:
1. place an empty `en.json` file with an empty JSON object in there... that way the fallback happens... or 
2. reverse it... so `en.json` is still populated and `en-US.json` is an empty JSON object... 

Either way works and gets us easily past the load... I'll add it into a review commit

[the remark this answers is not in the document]
priley86: woooot

priley86:
how long does this travis build go? I'm still clicking ["cookies"](https://orteil.dashnet.org/cookieclicker/) over here 😸

PR #81, opened 2019-08-07
feat(rhelGraphCard): add graph granularity dropdown

### LINKED ISSUE #6: Activate graph granularity dropdown for Daily, Weekly, and Monthly
## Feature description

As a customer I'd like the ability to select a granularity range and have the graph update accordingly.

## Acceptance Criteria
- [x] The graph dropdown list should contain the follow selections
   - **Daily** this represents an inclusive prior 30 day range (keep the last 30 days logic)
    - What to expect... expect daily to hard cutoff
   - **Weekly** this represents an inclusive prior 12 week range, in addition to the current week/day the user is within... "a bakers dozen"
      - What to expect... we define a week as Sunday to Sunday
      - What to expect... end date goes towards "a total for the week at that particular point in time"
   - **Monthly** this represents an inclusive prior 12 month range
      - What to expect... end date goes towards "a total for the week at that particular point in time"
   - **Quarterly** (TBD UI/UX still being worked) this represents 12 month range divided by 4. It should extend towards a 4 year or 16 quarter range.
      - What to expect... end date goes towards "a total for the week at that particular point in time"
- [x] The dropdown list should return a specific onSelect output that can be used across multiple component implementations
- [x] Graph granularity should be tracked at the Redux state level
- [ ] Graph display should have a y-axis that displays the year only when a result set spans multiple years
- [ ] Add additional mock/fixture service responses for local debugging
- [ ] Unit tests should be expanded to confirm predictable granularity graph result sets

## Storyboard
Screen Shot 2019-08-06 at 4 50 13 PM

## Assumptions & Questions
- CONFIRMED, the dropdown should also update the list view, confirm with @bclarhk
   - the graph dropdown **will NOT** update the list view, given the current UX direction
- Can we get a "started monitoring" date from the API that represents a kind of "start date", confirm with API team

priley86:
## What's included

* Adds the PF4 Select
* Starts the phase2 date range logic for RHEL Graph Card for given date ranges
* Modifies the date helpers to support weekly, monthly, quarterly
* Modifies the chart domain calculation logic: TLDR - this is currently causing problems. I believe the easiest solution would be to adopt the `tickValues` and the changes proposed in #70. This will force axis values to behave responsively without our CSS. Seems to be easier path at the moment.

## Example

Screen Shot 2019-08-07 at 3 54 14 PM

## Updates issue/story

Closes #6 #88


src/components/rhelGraphCard/rhelGraphCard.js
```diff
@@ -46,16 +52,15 @@ class RhelGraphCard extends React.Component {
...
-  };
+  onSelect = event => {
+    const { value } = event;
 
-  onToggle = dropdownIsOpen => {
-    this.setState({
-      dropdownIsOpen
-    });
+    if (event) {
+      store.dispatch({
```
priley86: redux it is! woot.

src/components/i18n/i18n.js
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import i18n from 'i18next';
+import i18next from 'i18next';
 import XHR from 'i18next-xhr-backend';
 import { initReactI18next } from 'react-i18next';
 import { helpers } from '../../common/helpers';
 
+const translate = (str, placeholder = null) =>
+  (!helpers.TEST_MODE && i18next.t(str, placeholder)) || helpers.noopTranslate(str, placeholder);
```
priley86: 👍

src/components/select/select.js
```diff
@@ -0,0 +1,216 @@
...
+        };
+
+        const mockEvent = {
+          ...mockTarget,
+          target: { ...mockTarget },
+          currentTarget: { ...mockTarget },
+          persist: helpers.noop
+        };
+
+        onSelect({ ...mockEvent }, optionsIndex, mockUpdatedOptions);
```
priley86: this is the kind of API response I would expect from a generic Dropdown/Select/OptionsMenu - if an object of options passed, it should be returned to the consumer as an object (that can they can access it again if needed).

Major kudos on this! Would like to see the same sort of behavior in PF some day! 👍

priley86:
This mostly looks good to me... just a couple of things I noticed in local mock mode.

Daily:
Screen Shot 2019-08-16 at 2 36 27 PM

Weekly:
Screen Shot 2019-08-16 at 2 36 36 PM

Monthly:
Screen Shot 2019-08-16 at 2 22 19 PM

Quarterly:
Screen Shot 2019-08-16 at 2 24 11 PM

Daily and Weekly seem to work great. 

So my UX questions for Monthly aspect - should you consider listing the year some place? Or is it obvious to assume it's the prior year?

For the Quarterly aspect - this is likely just bad mock data. We should update that to something more reasonable and then figure out how many days/tick values to actually render. I think Victory is trying to compensate for the offsets in an unexpected way.

cdcabrera:
Yup yup, it is most definetely the mock data not going back as far and the charting mechanics trying to compenstate. I'll go ahead and add these bits into this PR
1. Will add a few older data points in the mocks with the hopes it'll kick in what we do to fill in the gaps date wise
2. Add in the year into the quarterly view... there may also need to be something like this for cross year displays for all the ranges @bclarhk ?

bclarhk:
In the wires I put the year below January(s) in Monthly and Quarterly views
https://marvelapp.com/61fh138/screen/60020672

cdcabrera:
To see the below variations the mocks/fixtures have been updated with multiple `201` responses within `./src/services/rhelServices.js` 

It's a little manual, but change the `201` to a `200` and the correct response should populate per granularity filter, only one 200 is utilized...
```
 * @apiSuccessExample {json} DAILY, Success-Response:
 *     HTTP/1.1 200 OK
```

#### Daily
Screen Shot 2019-08-22 at 10 10 24 AM

#### Weekly
Screen Shot 2019-08-22 at 10 11 04 AM


#### Monthly
Screen Shot 2019-08-22 at 10 11 50 AM


#### Quarterly
Screen Shot 2019-08-22 at 10 12 20 AM

cdcabrera:
Things that have shifted
- chartArea component, we included the area chart conversion from issue #88 commit messaging will reflect once the review happens
- chartArea component, we auto adjust tick values
- chartArea component, we auto adjust the domain/range for x and y... y being based on a `max Y` number, and x being based on the length of the data utilized
- x no longer populates a value beyond a placeholder number that increments, the primary focus is on making sure the label displays per UX/UIs request

src/common/graphHelpers.js
```diff
@@ -1,45 +1,97 @@
...
+/**
+ * Return a time allotment based on granularity
+ *
+ * @param {string} granularity enum based on rhelApiTypes.RHSM_API_QUERY_GRANULARITY_TYPES
+ * @returns {string}
+ */
+const getGranularityDateType = granularity => {
+  switch (granularity) {
+    case GRANULARITY_TYPES.DAILY:
+      return 'days';
```
priley86: are these literals backend API constants? Maybe move this helper function into `rhelApiTypes` file instead of graphHelpers? Not really sure, just seems different.

src/common/graphHelpers.js
```diff
@@ -50,154 +102,113 @@ const getLabel = ({ data, previousData, formattedDate, label, previousLabel }) =
...
-    if (values[formattedDate]) {
-      chartData.push({
-        x: values[formattedDate].x,
-        y: values[formattedDate].y,
-        label: updatedLabel
-      });
+    const isNewYear = i !== 0 && year !== previousYear;
+    let formattedDate;
+
+    if (granularityType === 'months') {
```
priley86: constant for `months`?

src/common/graphHelpers.js
cdcabrera: so the rhelApiTypes are meant to be static... this is geared towards a specific date type from moment

src/common/graphHelpers.js
cdcabrera: same as above, these are from the `getGranularityDateType` aimed specifically at moment

src/common/graphHelpers.js
cdcabrera: if this type switch is used beyond the immediate helper, then willing to reconsidering moving it within a type file, or within the `dateHelpers.js`

src/common/__tests__/__snapshots__/graphHelpers.test.js.snap
```diff
@@ -1,223 +1,261 @@
...
+    "label": "t('curiosity-graph.tooltipLabel')",
+    "previousLabel": "t('curiosity-graph.tooltipPreviousLabelWeekly')",
+  },
+}
+`;
+
+exports[`GraphHelpers should return a x axis tick increment based on granularity: x axis tick increment 1`] = `
+Object {
+  "daily": 5,
+  "monthly": 1,
```
priley86: this is really nice! Custom increments by granularity...

These settings seem to be OK.. I would possibly consider increasing monthly to 2 though for better mobile spacing... it's legible though.
Screen Shot 2019-08-22 at 10 59 54 AM

cdcabrera: Debating on whether to iterate on this facet. Part of the graph helpers solution is temporary, would have to start tracking tick increment against label/year display to handle this request

cdcabrera: - pushed an update so the tick increment affects the displayed year
- pushed an update so months are now every 2

Screen Shot 2019-08-22 at 1 14 55 PM

bclarhk: Lookin good!
I think the colors/styling might get tweaked but we can merge that bit in later.

priley86: awesome!

PR #113, opened 2019-10-16
fix(graphHelpers,chartArea): issues/92 tooltip labels

### LINKED ISSUE #92: Discreet data points need alternative tooltip labels
## Description
As an end user I would like to not be confused between seeing a "zero" data point vs a "no data" point.

## Acceptance Criteria
- [x] discreet data points that are filler should have tooltips that read "no data", NOT "0 sockets" or "date"
- [x] tooltips for the initial graph point that displays should not highlight the previous data point by calling out the positive difference. It should be ignored and only show the x and y axis increments
- [x] a "0" y value, or discreet data point filler, after a positive y value should have a tooltip that reads "no data". Do NOT display the "negative difference" or "date"

## Assumptions and Questions
- The RHSM API still plans to return "0", or a discreet data point for nonexistent data. There are also times when "0" will be a valid value. In addition the API is planning to return a boolean that makes a distinction between an actual "0" vs filler

cdcabrera:
## What's included

- fix(graphHelpers,chartArea): issues/92 tooltip labels
   * graphHelpers, fix the label display, added No Data
   * chartArea, fix tooltip positioning
   * locales, add in No Data
- fix(rhelGraphCard,graphHelpers): correct y axis ticks
   * y axis ticks were being rounded incorrectly instead of abbreviated
   * updated package and unit test to confirm abbreviation continues

### Proxy run check
1. `$ yarn start:proxy`
1. hover over the graph to see the improved tooltips

## Example

Oct-16-2019 19-11-57

Oct-17-2019 11-39-26

## Updates issue/story

closes #92


cdcabrera:
Correction for y axis ticks applied, should now appear as
  
Screen Shot 2019-10-17 at 2 22 39 PM

cdcabrera:
Updated tooltip, and expanded testing scenarios for when there isn't reporting data but there IS capacity/threshold

Oct-17-2019 15-31-45

PR #196, opened 2020-02-06
fix(authentication): issues/190 relax redirect on certain users

### LINKED ISSUE #190: Relax platform modal redirect for users missing account numbers
## Description
As an end user the redirect modal is lacking information that I need to continue my process in viewing Subscription Watch. I'd like a relaxed platform modal redirect for users missing account numbers and to instead see the "tour" or "soon" empty state view.

## Acceptance Criteria
- [x] Users with missing account numbers should no longer receive the platform modal, but instead be redirected towards the soon/tour
- [x] Confirmation should be given by a UX/UI team member

## Assumptions and Questions
- Fastest solution is to relocate the redirect logic and have the end user redirected towards the soon/tour view
- Changing the content of the modal is the more intensive solution. May need to open a PR on the chroming repo directly to aid in getting the change pushed. It is unclear if this is Pendo related or a simple routed modal component based on query/search.
- Additional ask for turning off the toast notifications specific to the `soon`/`tour` view. 
   - Unfortunately the notifications are all or nothing, and in this scenario in order for the redirect to happen we first have to hit the API and receive the `403` which in turn generates the toast notifications and redirects.
   - Compromise solution? Once the platform modal redirect is turned off both malformed users and `403` responses will redirect towards the `soon`/`tour`. There will be no visual distinction between what error has happened. 
      - As a developer being randomly asked why something broke and given only a screenshot we're recommending a compromise solution of a single toast notification with approved copy, and possibly some copy indicator highlighting why the redirect happened. It is understood that merely redirecting towards the `soon`/`tour` should normally be that indicator. 
   - Another compromise solution, copy adjustments to the `soon`/`tour` highlighting the error.

cdcabrera:
## What's included

- fix(authentication): issues/190 relax redirect on certain users
   * combines the soon and redirect logic for malformed users

### Notes

- ~Need some UI/UX design input around toast notifications.~ Still need UI/UX input, however, after re-reviewing the toast notification options turning them off again appears to provide a smoother UI/UX experience.
   - ~Given we re-activated the toast notifications during development, they have become useful for being able to tell visually what form of error is taking place. Now that we're removing one of the distinctions in redirects and only using the `tour/soon` it would be helpful, as a developer, to maintain some minimal form of toast notification. Whether that was...~
      - All errors receive a toast notification with the error number and a brief message, possibly tailored
      - Or we only show a single toast in the specific instance of a user being malformed
- **The tour button needs to be confirmed as working, or updated, at the Pendo level with this change. The redirect updates replace the view component in place and does not necessarily update the browser route.**

### Local run check
1. `$ yarn start`
1. In order to emulate a quick `4XX` level error we'll need to update the mock server
    - Edit line number `42` in `/src/services/rhsmServices.js` ... change the `200` to a `401` and let the page refresh
    - Note: The made up status of `418` can not be emulated this way. It involves purposefully breaking line `10` around the `getUser` method in the `platformServices.js` file

#### Single toast notification from API error response
Screen Shot 2020-02-13 at 10 15 40 AM (2)

 #### Single toast notification for malformed users, needs copy adjustment
Screen Shot 2020-02-06 at 4 25 30 PM (3)

#### Single toast notification for general 4XX level http status (does not include 5XX status), needs copy adjustment
  
Screen Shot 2020-02-06 at 4 45 53 PM (3)

## Updates issue/story

Close #190


cdcabrera:
@jeperry @bclarhk 

Due to the routing and redirect updates, once this is within the CI/dev environment the primary tour button needs to be confirmed as both
- functioning
- and that the link that is dynamically applied **DOES NOT** use the browser url/route/path to determine when the link is applied. Because of updates this method of applying the link will break. 

As a helpful addition we added the CSS className `uiux-button__tour` to the primary button...

Screen Shot 2020-02-13 at 4 22 35 PM

jeperry:
@cdcabrera whats the best way to test this now?

PR #205, opened 2020-02-13
feat(openshiftView): issues/185 add cores sockets selector

### LINKED ISSUE #185: Add a unit switcher for the OpenShift product view graph
## Description
As an end user I would like to be able to switch the units displayed, sockets and cores, on the OpenShift product view graph.

## Acceptance Criteria
- [x] Redux state layer should consume and accept "switcher" broadcasts
   - Should include using base view/route to scope state object, relates to #179 
- [x] A unit switcher should be displayed on the OpenShift view graph
   - Should contain a Sockets option
   - Should contain a Cores option

## Assumptions and Questions
- UI/UX design work should be happening parallel
   - Units being switched are sockets and cores. "Cloud Meter style" was used as a reference. UI/UX confirmation needed first since the graphs have the ability to simply show both values at the same time.
- Initial GUI dev should include laying out the state layer, consuming and updating
- First round may be exactly as project management has requested... "Cloud Meter" like

Here's the Insta-design
image

cdcabrera:
## What's included

- feat(openshiftView): issues/185 add cores sockets selector
   * locales, add generic cardHeading string
   * graphCard, allow children to populate card actions
   * openshiftView, add cores sockets select list for graph filtering
   * rhelView, code style consistency against openshiftView

### Proxy run check
1. `$ yarn start:proxy`
1. log in with a user that has related data
1. navigate to the Openshift path and confirm the select list filters the displayed graph data

## Example

...

## Updates issue/story

Close #185

### DIFF
--- a/public/locales/en-US.json
+++ b/public/locales/en-US.json
@@ -5,6 +5,7 @@
     "authorizedCopy": "You do not have permission to access reporting. Contact your administrator."
   },
   "curiosity-graph": {
+    "cardHeading": "CPU usage",
     "coresHeading": "CPU core usage",
     "socketsHeading": "CPU socket usage",
     "dropdownDaily": "Daily",
--- a/src/components/graphCard/graphCard.js
+++ b/src/components/graphCard/graphCard.js
@@ -128,7 +128,7 @@ class GraphCard extends React.Component {
   }
 
   render() {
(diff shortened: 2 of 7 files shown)


bclarhk:
LGTM!

Issue #43, opened 2019-07-19
Consolidate, or colocate, and/or abstract `window.insights` global refs


cdcabrera:
## Description
As a developer it would be helpful to consolidate/colocate, and/or abstract `window.insights` global refs. Possibly encase it alongside `authorizeUser` within the service layer

## Acceptance Criteria
- [x] `window.insights` global refs should be within a central set of "platform services"

## Assumptions and Questions
- for consistency may be helpful to structure the responses similar to service calls, i.e. the "getUser" platform method, see #167 
- relates to #179

PR #209, opened 2020-02-19
fix(userServices): issues/43 catch empty response objects

cdcabrera:
## What's included

- fix(userServices): issues/43 catch empty response objects
   * expand generic auth response check, object and has keys

### Notes

- Catches a scenario where the original user data response check would pass a valid empty object.

### Environment check
1. Log into CI/QA with a malformed user that returns an empty object response (typically associated with a lacking account number)
1. Navigate to the Subscription Watch app
1. You should be redirected towards the "Tour View"

## Example

...

## Updates issue/story

Updates #43


cdcabrera:
Travis is acting weird on this PR closing and reopening

PR #211, opened 2020-02-25
fix(tourView): issues/190 copy update for malformed users

cdcabrera:
## What's included

- fix(tourView): issues/190 copy update for malformed users
   * router, test update for connected tourView
   * tourView, add session state, conditional display for malformed users
   * i18n, sentence break out for tour, test update

### Notes

- this applies an adjustment, per design, to the Tour View @rblackbu 
   - remove a sentence from the first paragraph with "tour" copy
   - remove the "tour" primary button and move the "learn more" into its place

### Local run check
1. `$ yarn start`
1. comment out the line of code, like below, at `line 17 ./src/services/userServices.js`
   ```
      if (_isPlainObject(userData) && Object.keys(userData).length) {
        // return Promise.resolve({ data: userData, message, status: 200 });
      }
   ```
1. observe the changes similar to the attached screenshot

### Proxy run check
1. `$ yarn start:proxy`
1. login with a "malformed" user
1. navigate to the tour view, observe the updates associated with the attached screenshot 

## Example

Screen Shot 2020-02-25 at 12 46 16 PM

## Updates issue/story

#190


cdcabrera:
@jeperry @bclarhk  @rblackbu

Per discussion it appears we/I can throw an update against the platform redirect modal. The code is located here 
- https://github.com/RedHatInsights/landing-page-frontend/blob/da7548c4c50d8c2bf0c7b799f7cb173d62a45983/src/consts/unentitled.js#L113-L124

This may be the best long-term solution. I'll reach out to the platform team to confirm us opening a change is acceptable.

mbpierce:
This works for me 
Let me know when it makes it in the big MR (unless I need to do something before then)

mbpierce:
Aaactually...
I realized that I didn't actually try and take the tour (which is a pendo functionality, I think). So I just went to click on the tour button and it didn't do anything. (I was logged in as a malformed user) The `Learn more` link under the button and to the left took me back to the login page, I assume because I was a bogus user in the system).

Anyway, if there's some other sort of malformed user I should be checking this with, let me know those details...

jeperry:
@mbpierce any Pendo-driven function wont work if the chrome shows you as not logged in. In the case of malformed users, you will see in the top right of the masthead the button 'Log in' instead of your login name and other menus that you should see. So as far as Pendo is concerned, it is the expected behavior that the tour link would not work as long as the chrome 'thinks' you are anonymous.

cdcabrera:
@bclarhk the updated sentence spacing on the "tour button" version screenshot
    
Screen Shot 2020-02-25 at 3 07 01 PM

mbpierce:
This looks good to me.

Issue #240, opened 2020-03-31
Add user account "opt in" form


cdcabrera:
## Description
As an end user I'd like to be able to easily opt into viewing my subscription data.

## Acceptance Criteria
- [x] Design, storyboard
- [x] Dev, Implementation should follow UI/UX storyboard
   - [x] Services, Redux breakout
   - [x] Card view component
   - [x] Tour view vs the OptIn form

### Extended Acceptance Criteria
As of 20200413, April 13th
- [x] Design, Storyboard updated
- [x] Dev, All users should now have "opt-in" button access rights
- [x] Dev, Remove the "[contact your admin]" copy/view
- [x] Dev, apply UI/UX copy updates
- [x] Dev, confirm API updates are released

## Assumptions and Questions
- Design & Dev, UX determination around the Tour view vs the OptIn form
- 20200413, admins are no longer the limiting factor for who receives the "opt-in" form. 
@rblackbu

rblackbu:
See pages starting after pg 3 https://marvelapp.com/114632g9

PR #256, opened 2020-04-15
fix(optinView): assume opt-in is activated on 4XX error

cdcabrera:
## What's included

- fix(redux): rename use of http errorStatus to just status 
   * redux, scenarios that involve using status for both success, fail in
- fix(optinView): assume opt-in is activated on 4XX error
   * optinView, check for 403 status, if not, assume opt-in is activated
- fix(helpers): access to an app specific window object 
   * browserExpose, allow GUI functionality in local browser dev console
- fix(services): apply consistent exports for config

### Notes

- On any 4XX error, NOT a 403, assume a user has access
- ~OPTIONAL ADDITION - on error confirm error code "SUBSCRIPTIONS1003" as a "user is opted in"~ Holding on this facet, something to iterate on, see... #260 
- Covers a one-off case of users navigating directly to the `/subscriptions/optin` route. In that scenario assume a user who is able to display said route can not "resubmit" the optin form.

1. `$ yarn test`

## Example

...

## Updates issue/story

#240


cdcabrera:
Per the API team start using an error code of 
```
{
  "errors": [
    {
      "status": "string",
      "code": "SUBSCRIPTIONS1003",
      "title": "string",
      "detail": "string"
    }
  ]
}
```

from ... https://github.com/RedHatInsights/rhsm-subscriptions/blob/5080abd485015029179b4d4284834eec34f247d9/src/main/java/org/candlepin/subscriptions/exception/ErrorCode.java#L53

Issue #320, opened 2020-06-16
Implement platform component for an application level maintenance mode


cdcabrera:
## Description
As a designer I should investigate a pf based solution that compliments the Platform level maintenance messaging to avoid overlap and confusion.

As a developer I need to develop a solution for application level maintenance mode for scenarios where coordination between API, Platform, and GUI can happen "awkwardly"

As an end-user it would be helpful to understand when the application is not-functioning to avoid confusion and help me not open issues against non-existent problems.

Maintenance notification
https://docs.google.com/document/d/1VLs7vFczWUzyIpH6EUsTEpJugDsjeuh4a_azs6IJbC0/edit

SVG
https://pnt.redhat.com/pnt/p-15530715/Icon-Red_Hat-...Black-RGB.svg

Component:
https://github.com/RedHatInsights/frontend-components/blob/master/packages/components/src/Components/Maintenance/Maintenance.js

## Acceptance Criteria
- [x] Determine the best location to place the maintenance mode component. 
    - authentication, or as a fallback the app component.
    - determine if locale strings is an option. i.e. moving strings to our en-US.json if unavailable
    - if strings are required work with UI/UX, the API team, and CCS to determine required copy
- [x] Currently there is no API endpoint associated with activating this mode
   - confirm/follow up with API team on the related Jira task
   - if the API is unavailable determine if a temporary disable would benefit us, similar to the existing `disables` located in our core dotenv `.env`

## Assumptions and Questions
- Will copy will be generated from the API, or will the API continue with error codes?
- The Auth component brings everything together, however the `app` component may be a better location. Reasons for consideration should include an "application level" maintenance mode would still give us access to RBAC and general platform permissions
- Unclear how often this will happen. But given the recent RBAC changes it would helpful to an end user, and help avoid unnecessary bug filings
- Existing design, and a plan at the Platform level to make this a shared component most likely through the Frontend-Components repo.
![image](https://user-images.githubusercontent.com/32572198/87810825-98bea000-c82b-11ea-8f09-393f21c05e5d.png)

cdcabrera:
As of 20200717 @rblablu hourglass icon research. if you can locate said icon, super helpful!

rblackbu:
Replace icon with the latest 'hourglass' style. See the attached reference. 

Download brand icon library here --> [
[...]

PR #379, opened 2020-08-18
feat(maintenanceView): issues/320 application level maintenance mode

zanewoodfin:
## What's included

- feat(maintenanceView): issues/320 application level maintenance mode
  - maintenanceView, create page using Maintenance component from frontend-components library
  - router, redirect to maintenanceView when `REACT_APP_UI_DISABLED=true`

### Local run check
1. set `REACT_APP_UI_DISABLED=true`
1. `$ yarn start`
1. visit any route
1. confirm you are redirected to the new maintenance page

## Example

maintenance_mode_draft1

## Updates issue/story
Updates #320


src/components/messageView/messageView.js
```diff
@@ -11,33 +11,39 @@ import { helpers } from '../../common';
...
   <PageLayout>
     <PageHeader>{pageTitle || helpers.UI_DISPLAY_NAME}</PageHeader>
-    <EmptyState variant={EmptyStateVariant.full} className="fadein">
-      {icon && <EmptyStateIcon icon={icon} />}
-      {title && (
-        <Title headingLevel="h2" size="lg">
-          {title}
-        </Title>
+    <PageSection>
+      {(children && children) || (
```
cdcabrera, line 25: required: 
1. So little known, and not every repo has this, but we have access to the nullish coalescing operator through our build setup. It should help clean up this syntax. Instead it'd be...
   ```
   { children ?? (
     <EmptyS...
   ```
1. And it would probably be a good idea to drop in a quick unit snapshot test around `children`

src/components/messageView/messageView.js
```diff
@@ -11,33 +11,39 @@ import { helpers } from '../../common';
...
-    </EmptyState>
+    </PageSection>
   </PageLayout>
 );
 
 /**
  * Prop types.
  *
- * @type {{icon: Node|Function, message: string, title: string}}
+ * @type {{children: Node, icon: Node|Function, message: string, pageTitle: string, title: string}}
```
cdcabrera, line 43: good catch  on the missing annotation!

src/components/messageView/messageView.js
zanewoodfin: Oh man what I did with `(children && children) ||` was totally redundant.  Should've just been `children ||`.

cdcabrera: that totes works too!

Issue #365, opened 2020-08-04
Research and implement alternate chart colors to improve contrast


cdcabrera:
## Description

As a UI/UX designer I need to determine higher contrasting colors to display on the chart/graph.

As a developer I need to implement the recommended color updates.

## Acceptance Criteria
- [x] UI/UX color research
- [x] Stroke width updates
   - Applies across product views
- ~Dev implementation of colors~ revisit
   - ~Openshift view updated~
   - ~RHEL view updated~

## Assumptions and Questions
- color updates are handled in the default props of the OpenShift and RHEL views.
- will this sorting order issue affect the color scheme, see #329

PR #392, opened 2020-08-26
fix(graphCard,chartArea): issues/365 increase stroke width

zanewoodfin:
## What's included

- fix(graphCard,chartArea): issues/365 increase stroke width

### Notes
Instead of updating colors, I've increased the stroke width.  I updated the primary stroke width from 1 (default) to 2.  I also updated the dashed line stroke width from 2.5 to 3, so that it would still be visibly thicker than the other lines.

1. `$ yarn test`

## Example

strokeWidth2

## Updates issue/story
Updates #365

### DIFF
--- a/src/components/chartArea/chartArea.js
+++ b/src/components/chartArea/chartArea.js
@@ -477,6 +477,10 @@ class ChartArea extends React.Component {
         dataColorStroke.data.stroke = dataSet.stroke;
       }
 
+      if (dataSet.strokeWidth) {
+        dataColorStroke.data.strokeWidth = dataSet.strokeWidth;
+      }
+
       return (
         <PfChartArea
           animate={dataSet.animate || false}
--- a/src/components/graphCard/graphCard.js
+++ b/src/components/graphCard/graphCard.js
@@ -118,6 +118,7 @@ class GraphCard extends React.Component {
             duration: 250,
             onLoad: { duration: 250 }
           },
+          strokeWidth: 2,
           isStacked: !/^threshold/.test(key),
           isThreshold: /^threshold/.test(key)
         };
@@ -129,7 +130,7 @@ class GraphCard extends React.Component {
           };
           tempFiltered.stroke = chartColorGreenDark.value;
           tempFiltered.strokeDasharray = '4,3';
-          tempFiltered.strokeWidth = 2.5;
+          tempFiltered.strokeWidth = 3;
         }
 
         return tempFiltered;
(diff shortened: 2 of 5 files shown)


src/components/chartArea/chartArea.js
```diff
@@ -477,6 +477,10 @@ class ChartArea extends React.Component {
         dataColorStroke.data.stroke = dataSet.stroke;
       }
 
+      if (dataSet.strokeWidth) {
+        dataColorStroke.data.strokeWidth = dataSet.strokeWidth;
+      }
+
```
cdcabrera, line 483: required:
add a minor adjustment to the unit test for `chartArea` so we can see this come through

the `... render basic data` test should do it since it's already using a "strokeWidth" for the threshold
-  https://github.com/RedHatInsights/curiosity-frontend/blob/ci/src/components/chartArea/__tests__/chartArea.test.js#L13-L55

```
...
          stroke: '#lorem',
          strokeWidth: 2
        },
...
          strokeDasharray: '4,3',
          strokeWidth: 3
        }
```

src/components/graphCard/graphCard.js
```diff
@@ -129,7 +130,7 @@ class GraphCard extends React.Component {
           };
           tempFiltered.stroke = chartColorGreenDark.value;
           tempFiltered.strokeDasharray = '4,3';
-          tempFiltered.strokeWidth = 2.5;
+          tempFiltered.strokeWidth = 3;
```
cdcabrera, line 133: Optional:
The history behind these changes is best described with the phrase "3rd time is a charm"

Since we're updating the Patternfly Charts base settings, and we're doing it again, very tempted to request you put a `ToDo:` annotation above this class method that says something along the lines of "[minimums are good, but we've changed this multiple times now. Styling for graphs/charts in the future, should be moved to the filters in the default props of the product views]" 

```
/**
   * ToDo: future consideration, maintain style minimums as default props in graphCard, move chartArea customizations to the filters used on product view configs.
   * FixMe: custom use of dash...
```

rblackbu:
Looks good

src/components/graphCard/graphCard.js
zanewoodfin: I tried to figure out what you wanted me to do here, but I think I'm just not sure what the goal is.  Are you suggesting that we move the width values to props, so that we could change the widths in the future at the view level?  If so, I'm not sure why we would ever want to do that.  Shouldn't all charts of this type have the same widths?

If anything, I felt like we should be setting the widths deeper in, like in the ChartArea component.  I only put the changes here because that's where things were already being set.

rblackbu: +1 to all same widths

cdcabrera: The goal was to consider it more of config based setup without specific settings being nested within a class method. There are settings that could be offset into the helpers, similar to the xAxis, yAxis, like "threshold" and "stacked". A chartArea helper could also be an answer, similar to your mention. 

We do at some point need to restructure the entire `chartArea.js` since there have been updates to the PF components that might, and I stress the might part, might help us cut code.

cdcabrera: We'll opening a follow up issue to help think about restructure.

PR #429, opened 2020-09-17
chore(testing): testing GitHub workflow

cdcabrera:
## What's included

- chore(testing): testing GitHub workflow

## Example

...

## Updates issue/story

Ongoing


cdcabrera:
Successful test of the GitHub PR workflow vs Travis CI. Half the time. 

It does appear like we'll need to keep Travis alongside the GitHub workflow for PRs until the related workflow commit makes its way through all the branches. However we can make it so only the GitHub workflow is required.

There may be a continuing side-effect. There appears to be an occasional glitch either caused by Travis or CodeCov, we can't tell, that occasionally means CodeCov may not post coverage back in the PR comments. This may continue to happen until Travis is removed, or not, but if it does continue that is no different than what normally would happen if Travis was running standalone.

Issue #403, opened 2020-09-02
On inventory page update scroll position should not reset


cdcabrera:
## Description
Scroll position is reset towards the top of the interface when paging is implemented. This reset is due to the immediate page height update that happens because inventory enters a "pending" state. **The priority on this aspect is low**. While annoying it does not prevent access to data.

In addition updating the granularity appears to cause the expandable "guests" section collapse. These are 2 different queries and granularity updates should not affect the inventory display.

### Expected behavior

Primary behavior
On paging update I should not have to re-scroll to have the inventory centered in my display/browser

Secondary behavior
There is a secondary aspect that should not collapse the expandable guests display when granularity is updated.

#### Steps to reproduce

Scroll Issue
1. Open the Subs Watch interface
1. Scroll down to the inventory listing
1. Use paging to select another page
1. Confirm the view scroll position resets

Collapse Issue
1. Open the Subs Watch interface
1. Scroll down to the inventory listing
1. Expand a guests display
1. Update the graph granularity
1. Confirm the expanded guests display collapsed

### Additional context, version info

- possible solutions
   - general min height adjustment
   - dynamic min height adjustment after each render
   - storing scroll position

PR #439, opened 2020-09-26
fix(inventoryList,graphCard): issues/403 set minheight on update

cdcabrera:
## What's included

- feat(minHeight): issues/403 set min-height, avoid page jumps
- fix(inventoryList,graphCard): issues/403 set minheight on update

### Notes

- Applies a minHeight component to help resolve API page load jumps
- Column shifts are a potential subsequent fix

### Proxy run check
1. `$ yarn start:proxy`
1. login with an account that has numerous inventory items, then
   1. navigate to the next page in the inventory list, the page should no longer shift in it's over all height

## Example

Sep-30-2020 14-36-28

## Updates issue/story

#403

### DIFF
--- a/src/components/graphCard/__tests__/__snapshots__/graphCard.test.js.snap
+++ b/src/components/graphCard/__tests__/__snapshots__/graphCard.test.js.snap
@@ -69,80 +69,90 @@ exports[`GraphCard Component should render a non-connected component: non-connec
 <Card
   className="curiosity-usage-graph"
 >
-  <CardHeader>
-    <CardTitle>
-      <Title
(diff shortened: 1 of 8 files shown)


ntkathole:
@cdcabrera Fix is working fine, I just want to let you know that if page limit is large and number of hosts on page is very small, it create long blank page on UI. Not sure if somehow we can/should improve on it.

To reproduce it, have 101 hosts in account, set limit to 100 and navigate to next page. 

huge_blank_page

cdcabrera:
height fix is here, https://github.com/RedHatInsights/curiosity-frontend/pull/447

PR #447, opened 2020-10-07
fix(inventoryList): issues/403 minHeight update on perpage

cdcabrera:
## What's included

- fix(inventoryList): issues/403 minHeight update on perpage
- fix(minHeight): issues/403 resize events reset height
- fix(inventoryList): activate paging for smaller screen sizes
- fix(chartArea): resize observer test addition

### Notes
- Corrects min-height issue when both changing "per page" on paging and adjusting the browser size
   - @ntkathole 
- Re-activates the paging display CSS for smaller screens. This also appears to affect the primary inventory, and appears Patternfly related, https://www.patternfly.org/v4/components/pagination
- Expand chartArea unit test

### Proxy run check
1. `$ yarn start:proxy`
1. choose an account with a large inventory and select "100" entries per-page
1. then navigate to the next page with 100+- entries
1. then select 10 entries per-page, the min-height of the page should re-adjust

##### Huge blank page
Huge blank page example
##### Missing paging on smaller screens
Screen Shot 2020-10-07 at 4 59 49 PM

##### UPDATED paging on smaller screens
Screen Shot 2020-10-07 at 4 58 26 PM

## Updates issue/story

#403

### DIFF
--- a/src/components/chartArea/__tests__/chartArea.test.js
+++ b/src/components/chartArea/__tests__/chartArea.test.js
@@ -381,6 +381,22 @@ describe('ChartArea Component', () => {
     expect(component.state().chartWidth).toEqual(1337);
   });
 
+  it('should attempt to handle a ResizeObserver', () => {
+    const observe = jest.fn();
+    const unobserve = jest.fn();
+
+    window.ResizeObserver = jest.fn().mockImplementation(() => ({
+      observe,
+      unobserve
+    }));
+
+    const component = mount(<ChartArea />);
+    expect(observe).toHaveBeenCalledTimes(1);
+
+    component.unmount();
+    expect(unobserve).toHaveBeenCalledTimes(1);
+  });
+
   it('should run componentWillUnmount method successfully', () => {
     const component = mount(<ChartArea />);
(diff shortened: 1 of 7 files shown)


ntkathole:
@cdcabrera table looks good but user could not change "per page" value for system table on smaller screen.
Screenshot from 2020-10-08 13-50-39

cdcabrera:
@ntkathole 
1. So the "per page" selector being hidden is on purpose, that's expected per design
1. The filters being awkward is mostly expected, not working is unexpected. https://www.patternfly.org/v4/components/toolbar#with-filters,
   - Everything appears to partially work as intended when compared against Patternfly. We'll have to catch up with design on why the Platform's inventory filters, at smaller screen sizes, don't fully match Patternfly, it may be a set of changes in the future.
   - The "partial" working seems to happen around selecting the first filter not showing up in the "clear all filters" area, a second filter selection seems to make it work, it may be a "refresh" issue. We'll have to investigate this by first updating the PF NPM resource since the Patternfly demo site appears to have their demo fully working. We don't directly update the "clear all filters" so we're unsure since it's supposed to be baked into the component. 
Oct-08-2020 10-06-04

Issue #468, opened 2020-10-21
Hypervisor should display as "hypervisor" or "virtual" infrastructure type in system table


ntkathole:
## Description
We consider tally of hypervisor in **virtual** bucket but system table shows **physical** . (Previously we used to show as **Hypervisor**)

**Reproducible only in CI-Beta.**

### Expected behavior
Hypervisor should show as "hypervisor" or "virtual" not physical.

#### Steps to reproduce
1. Have hypervisor in account
2. See system table

### Additional context, version info
![Screenshot from 2020-10-21 18-49-03](https://user-images.githubusercontent.com/16951913/96726269-68a90f00-13cf-11eb-8b0a-ad8da44c92a9.png)

![Screenshot from 2020-10-21 19-06-00](https://user-images.githubusercontent.com/16951913/96727273-8034c780-13d0-11eb-9d5c-34596e3b6b81.png)

rblackbu:
Hey Nikhil, we are taking a look at this now and will post a direction by the end of the day. Thanks!

cdcabrera:
As discussed and indicated by UX/UI, specifically regarding the "hypervisor" display

1. If it's a hypervisor, it’ll have a "chevron twistie" and a number of VMs bubble

And there are additional updates around the display meant to align to the graph and associated tooltips
1. the inventory column formally called "Infrastructure" is now called "Type
1. the guests inventory column formally called "Name" is now called "VM name"

cdcabrera:
Updated discussion with UX/UI and API, as an agreed upon iterative "next steps" the GUI will be
1. Updated to use the Measurement Type column from the API
   - The API will review replacing the Cloud Providers currently in the response to simple be "CLOUD"
   - The API will review replacing the Hypervisor currently in the response with the word "VIRTUAL"
   - The GUI will display the word "Hypervisor" until the API is updated
   - The GUI will attempt to temporarily replace the Cloud Providers coming through in the response with the phrase "Public Cloud"... we may simply pass the data for display
1. The acronym "VM" will be replaced in two locations to say "Guests [TOTAL]" and "Guest name" similar/same to the "pre" VM display
1. The GUI will hold on displaying the column header change "Subscribed sockets"

@bclarhk @rblackbu  @jlprevatt

bclarhk:
This looks like a good iterative path forward.

cdcabrera:
Per the requested updates, see PR 

Primary gotcha is lack of sorting on "measurement type"
[...]

Issue #389, opened 2020-08-24
Differentiate between public cloud and virtual system in system table


ntkathole:
## Description
Currently public cloud system is also shown as virtual in infrastructure column. This may create confusion when end user tried to calculate tally based on system table. Graph shows public cloud system usage differently than virtual system usage and also API has  'measurement_type'  based on which public cloud systems can be differentiate, but there is no way in UI to differentiate between them.

**API :**
```
{'cores': 24,
           'display_name': '62078ddc-8bdb-49c9-93af-b48e0a00768f',
           'hardware_type': 'VIRTUAL',
           'insights_id': '7c2c5868-c11d-4601-9b70-a3da1205a6ba',
           'inventory_id': '62078ddc-8bdb-49c9-93af-b48e0a00768f',
           'last_seen': datetime.datetime(2020, 8, 24, 4, 32, 5, 262810, tzinfo=tzutc()),
           'measurement_type': 'AZURE',
           'number_of_guests': None,
           'sockets': 1,
           'subscription_manager_id': 'ccae7d2b-0ae1-425e-8629-84348f034f3e'},
```

**GUI :**
![Screenshot from 2020-08-24 10-11-30](https://user-images.githubusercontent.com/16951913/91004585-3b6bfa80-e5f2-11ea-98b7-a102aa6c7c37.png)

### Expected behavior
Infrastructure type should be public cloud (not virtual) for public cloud usage.

#### Steps to reproduce
1. Have a public cloud usage.
2. See system table shows all systems "virtual"

### Additional context, version info

Related to #10 #393

cdcabrera:
As of 20201005 API is planning to push up to dev this week

cdcabrera:
@ntkathole some of the updates on the API side made it to CI we're aware that the translation strings aren't carrying over in that environment

![Screen Shot 2020-10-15 at 4 18 04 PM](https://user-images.githubusercontent.com/3761375/96181495-1749dd00-0f02-11eb-9a69-c86c09d6876d.png)

Appears instead of being called "virtual" they're now called "virtualized" =)

ntkathole:
Fixed! Now we have public cloud type in system table.

PR #469, opened 2020-10-21
fix(openshiftView,rhelView): issues/468 header names

cdcabrera:
## What's included

- fix(openshiftView,rhelView): issues/468 header names
- fix(helpers,testing): noopTranslate, helpful test snapshots

### Notes

- Header names, this is an iterative update aimed at expanding the helpfulness of the copy used for the headers
- `noopTranslate`, simple testing update to actually render out the JSON instead of displaying `[object Object]` in snapshots

### Proxy run check
1. `$ yarn start:proxy`
1. Locate an account with "VMs", expand the section, the new "VM name" should display

## Example

  
Screen Shot 2020-10-21 at 3 15 59 PM

## Updates issue/story

#468 
relates #389

### DIFF
--- a/src/components/toolbar/__tests__/__snapshots__/toolbar.test.js.snap
+++ b/src/components/toolbar/__tests__/__snapshots__/toolbar.test.js.snap
@@ -157,7 +157,7 @@ Array [
       Object {
         "data": Object {
           "clearFilters": Object {
-            "sla": null,
+            "usage": null,
           },
         },
         "type": "SET_QUERY_CLEAR",
@@ -312,7 +312,7 @@ Array [
       Object {
         "data": Object {
           "clearFilters": Object {
-            "sla": null,
+            "usage": null,
           },
         },
         "type": "SET_QUERY_CLEAR",
@@ -468,7 +468,7 @@ Array [
       Object {
         "data": Object {
           "clearFilters": Object {
-            "sla": null,
+            "usage": null,
           },
         },
         "type": "SET_QUERY_CLEAR",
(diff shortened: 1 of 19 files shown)


src/components/toolbar/__tests__/__snapshots__/toolbar.test.js.snap
```diff
@@ -157,7 +157,7 @@ Array [
       Object {
         "data": Object {
           "clearFilters": Object {
-            "sla": null,
+            "usage": null,
```
cdcabrera, line 160: required:
double check why this snapshot shifted, may have to do with the "default" on a switch

cdcabrera: It's related to `"type": "SET_QUERY_CLEAR"` and appears to be valid, it's unclear why it updated now, we'll note and keep an eye on it

cdcabrera: Figured it out... the issue was test specific not on the component or helpers, around line 83 https://github.com/RedHatInsights/curiosity-frontend/blob/ci/src/components/toolbar/__tests__/toolbar.test.js#L83

We were doing a "find" against a borked string... then we made more helpful snapshots

before
```
console.log src/components/toolbar/__tests__/toolbar.test.js:85
      CHECK >>> onSelect usage t(curiosity-toolbar.category, [object Object])
```
after
```
console.log src/components/toolbar/__tests__/toolbar.test.js:85
      CHECK >>> onSelect usage t(curiosity-toolbar.category, {"context":"usage"})
```
And everything started working... apparently the test was broken before =(

PR #472, opened 2020-10-23
feat(inventoryList,guestsList): issues/494 fixed column widths

cdcabrera:
## What's included

- feat(inventoryList,guestsList): issues/494 fixed column widths

### Notes

- Adds the ability to set fixed inventory column widths through config
   - Even though we're applying column widths there is still a potential shift in inventory paged displays that have/do not have expandable sections. The Patternfly expandable table rows add an additional table column that shift the view.
- Table skeleton should have fixed column widths applied too
- inventoryListHelpers changes are prep for #483 

### Proxy run check
1. `$ yarn start:proxy`
1. confirm during paging of the inventory display that columns maintain their width in a consistent manner. 

## Example

Screen Shot 2020-11-05 at 3 59 47 PM

Screen Shot 2020-11-06 at 12 22 01 PM

## Updates issue/story

#494 
relates #483

### DIFF
--- a/src/components/guestsList/__tests__/__snapshots__/guestsList.test.js.snap
+++ b/src/components/guestsList/__tests__/__snapshots__/guestsList.test.js.snap
@@ -19,7 +19,7 @@ exports[`GuestsList Component should handle multiple display states: fulfilled 1
       <Table
         ariaLabel={null}
         borders={false}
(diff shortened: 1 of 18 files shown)


cdcabrera (commented):
Rework tableSkeleton tests

Issue #473, opened 2020-10-23
Hide SAP/global filter from swatch


ntkathole:
## Description
Subscription watch page shows SAP/global filter to tag. I think it is not needed in subscription watch app.

Platform issue - https://issues.redhat.com/browse/RHCLOUD-10185 

### Expected behavior

```
Karel Hala added a comment - 33 minutes ago - edited
We've enabled global filter on all insights application and since subs watch is partially under insights bundle you are seing this. If you don't want to see it you have to disable this feature in your application https://github.com/RedHatInsights/insights-chrome#toggle-global-filter-on-certain-pages

EDIT: Just call `insights.chrome.hideGlobalFilter()` in your application when it first renders and you should be good go.
```

#### Steps to reproduce
1. Navigate to curiosity on QA/QA-beta

### Additional context, version info
![Screenshot from 2020-10-23 12-19-06](https://user-images.githubusercontent.com/16951913/96979630-cc9d1600-153c-11eb-8a14-dec525490206.png)

cdcabrera:
We'll go ahead and aim to align this to the push we were aiming for on 10/26. Our side of things may be delayed dependent on "deactivating" this piece

ntkathole:
**Exact steps to reproduce:**
1. Login with swatch-satellite-user to https://qa.cloud.redhat.com/
2. Navigate to subscription watch (Global filter will not be visible)
3. Navigate to "Red Hat Insights" using application switcher from left navigation panel. 
4. Select "Patch" from navigation.
5. Apply any tag from global filters.
6. Navigate to subscription watch using application switcher from left navigation panel.

![globfilter](https://user-images.githubusercontent.com/16951913/97144009-bafc7e00-1789-11eb-858a-05219e5ab9b5.gif)

cdcabrera:
Checked it out, currently still unable to recreate even with the steps listed. We'll look and see if some of the design team, @bclarhk @rblackbu ,  members can recreate this, wondering if browser is playing a part...


And in the spirit of Monday AM... we're once again seeing one of our older errors that we had squashed a ways back pop back up, something to do with how the redirect and `app name` are being handled on the Platform chrome side when the app is forced to redirect and the Platform dumps a user on ... `...com/subscriptions` 
- https://github.com/RedHatInsights/curiosity-frontend/issues/179

bclarhk:
I couldn't reproduce this on my Mac running Firefox 81.0.1

cdcabrera:
[...]

PR #474, opened 2020-10-23
fix(platformServices): issues/473 apply hideGlobalFilter

cdcabrera:
## What's included

- fix(platformServices): issues/473 apply hideGlobalFilter
   * authentication, hideGlobalFilter, open methods for testing
   * build, setupTests, expand scope of platform methods
   * platformServices, hideGlobalFilter platform chrome method
   * redux, actions, types, apply hideGlobalFilter

### Notes

- Disables the platform global filter
- Expands testing on the authentication component
- We're still unable to recreate this in the environment, we're working through emulation and memory on this one. We've implemented the Platform method similar to how we handled the other global methods. In the event this method fails, IN ENVIRONMENT, it should NOT take down the GUI, but will throw a console error message. Testing locally or through proxy will throw a giant error message as is normal.

### Proxy run check
1. make sure you enable Redux console messages, https://github.com/RedHatInsights/curiosity-frontend/blob/master/CONTRIBUTING.md#debugging-redux
1. `$ yarn start:proxy`
1. confirm that the Redux console messaging coming through either displays as a success or failure
   - success
        Screen Shot 2020-10-23 at 1 42 58 PM
   - failure if the method doesn't exist
      Screen Shot 2020-10-23 at 1 46 44 PM

   

## Example

...

## Updates issue/story

#473


ntkathole:
@cdcabrera Hey issue is still reproducible. Added exact steps to reproduce at https://github.com/RedHatInsights/curiosity-frontend/issues/473#issuecomment-716358495

Issue #502, opened 2020-11-10
Display messaging for a "has cloud meter data" flag


cdcabrera:
## Description
As a UX/UI person copy is needed to highlight when data discrepancies happen around Public Cloud on both the graph and inventory display

 ## Acceptance Criteria
- [x] Messaging is displayed on API condition
   - [x] QE confirmation, see #511 #512 - also relates indirectly to #499 as the basis for the #511 fix includes daily state handling
   - [x] ~CCS confirmation~ Fallback messaging implemented
   - [x] UX/UI confirmation

## Assumptions and Questions
- 2 ~potential~ flags from API ~, "mismatch" and "cloudigrade"~ coming through on the Tally response, https://github.com/RedHatInsights/rhsm-subscriptions/pull/218
   -  `has_cloudigrade_data`: true if cloudigrade data was included in the  snapshot, false otherwise.
   - `has_cloudigrade_mismatch`: true if there was cloudigrade data and the instance count differed from HBI data, false otherwise.
- Current plan, using the below screenshot
   - Maintain the close "X"
   - Use the "info" alert variant
   - If the link does not get generated plan on fallback copy
![Screen Shot 2020-11-11 at 1 41 35 PM](https://user-images.githubusercontent.com/3761375/98860179-46843900-2431-11eb-903b-5987151641e8.png)
- Link should be `https://access.redhat.com/subscription-watch-accuracy`

cdcabrera:
As of 20201130 updated messaging   ...
![Screen Shot 2020-11-30 at 3 50 22 PM](https://user-images.githubusercontent.com/3761375/100779952-2532bf00-33d7-11eb-90f4-e428435ec325.png)

PR #503, opened 2020-11-12
feat(bannerMessages): issues/502 cloud meter messaging 

cdcabrera:
## What's included

- feat(bannerMessages): issues/502 cloud meter messaging 
   * build, dotenv flag for recommendations link
   * locale, cloud meter related banner message strings
   * styling, banner messages spacing
   * appMessagesSelectors, apply tally cloud meter data, mismatch
   * bannerMessages, display banner messages
   * graphCardSelectors, expose cloud meter flags
   * rhelView, openshiftView, apply bannerMessages display
   * rhsmApiTypes, cloud meter data and mismatch flags
   * rhsmServices, mock update for cloud meter data
- feat(pageMessages): issues/502 section for messaging 
   * pageLayout, allow pageMessages
   * pageMessages, messaging display
   * pageToolbar, alignment with pageMessages

### Notes
@jlprevatt @ntkathole @bclarhk @rblackbu 
- The strings used for messaging are pre-approved, it is expected they be updated once CCS provides confirmation, 
- The link still needs to be created for docs, until it is the fallback message will display
- **Behavior**
   - The banner message area is applied to all products and ONLY displays if the `has_cloudigrade_mismatch` flag is `true` on any one data point within the Tally API endpoint
   - The banner may not initially display dependent on if the `has_cloudigrade_mismatch` flag is `false` across all data points on a "granularity". Subsequent granularity selections may fire the banner and display it
   - Once the banner is on REGARDLESS of whether a granularity DOES HAVE or DOES NOT HAVE the `has_cloudigrade_mismatch` flag, the banner will stay on.
      - Not implementing this behavior caused the banner to activate ON and OFF repeatedly when switch granularity
   - The banners have the ability to be "closed"... this is a temporary internal state and the banner will be reactivated when a user switches Products or architectures and the `has_cloudigrade_mismatch` flag is `true`

### Proxy run check
1. `$ yarn start:proxy`
1. log in with an account that has mismatched data from Cloud Meter, confirm the messaging displays

### With link
Screen Shot 2020-11-12 at 2 11 07 PM

### Without the link
Screen Shot 2020-11-12 at 2 10 19 PM

### Behavior
Includes closing the banner
Nov-12-2020 14-29-14

## Updates issue/story

#502


src/components/pageLayout/pageMessages.js
```diff
@@ -0,0 +1,36 @@
...
+ * @type {{children: Node, className: string}}
+ */
+PageMessages.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string
+};
+
+/**
+ * Default props.
+ */
```
cdcabrera, line 31: required: missing default props jsdoc type

src/components/pageLayout/pageToolbar.js
```diff
@@ -7,22 +7,30 @@ import { Section } from '@redhat-cloud-services/frontend-components/components/c
...
-  children: PropTypes.node.isRequired
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string
 };
 
 /**
  * Default props.
  */
-PageToolbar.defaultProps = {};
+PageToolbar.defaultProps = {
```
cdcabrera, line 32: required: missing default props jsdoc type

cdcabrera, line 32: required: missing default props jsdoc type

src/redux/selectors/__tests__/appMessagesSelectors.test.js
```diff
@@ -0,0 +1,215 @@
...
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.DATE]: '2018-07-04T00:00:00.000Z',
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.CORES]: 2,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.SOCKETS]: 2,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.HYPERVISOR_CORES]: 1,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.HYPERVISOR_SOCKETS]: 1,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.PHYSICAL_CORES]: 1,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.PHYSICAL_SOCKETS]: 1,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.HAS_DATA]: true,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.HAS_CLOUDIGRADE_DATA]: true,
+                    [rhsmApiTypes.RHSM_API_RESPONSE_PRODUCTS_DATA_TYPES.HAS_CLOUDIGRADE_MISMATCH]: true
```
cdcabrera, line 201: optional: consider flipping the `CLOUDIGRADE_MISMATCH` bool over to false here. the caching should show the same result as true from the original

src/redux/selectors/appMessagesSelectors.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @type {{appMessages: {cloudigradeMismatch: boolean}}}
+ */
+const selector = createSelector([statePropsFilter], data => {
+  const { viewId = null, productId = null, reportCapacity = {} } = data || {};
+  const appMessages = {
+    cloudigradeMismatch: false
+  };
+
+  const cache = (viewId && productId && selectorCache.data[`${viewId}_${productId}`]) || undefined;
```
cdcabrera, line 36: note: the cache is specific toward viewId and productId making the banner message shut down between architectures. if we want to say the banner is activated across a products architectures we should remove `...${productId}` from the cache key/string

cdcabrera (commented):
It was also noted there may be some iterative changes 
- making the Cloud Meter mismatch banner ONLY display using the most recent day/last 2 days. Unclear whether this would help reset the banner if the end user was able to correct their setup
- making the banner message remember it has been closed. multiple options/levels of remembering
   - at the Redux state level where it resets on product changes, but attempts to maintain being closed across architecture navigation
   - at a localStorage, sessionStorage level... sessionStorage resetting if the user closes the browser tab
   - at a cookie level, possibly set an expiration, resetting when a user resets their cookies

ntkathole:
Found blocker issue with backend https://issues.redhat.com/browse/ENT-3275 .. will verify this once it is resolved.

Issue #490, opened 2020-11-03
Pendo tour button should be actual component


cdcabrera:
## Description
As a developer I need to create a component that Pendo can access for the "Tour" label button.

## Acceptance Criteria

- [x] button component for Pendo created
- [x] UX signoff on Pendo access

## Assumptions and Questions
- button component can be nested inside the page header component, but alternatives can be considered
- the original button had z-index and absolute position issues
- Contact Margot M. after completion for Pendo sync
- Should behave similar to the OptIn button... just a button that lets Pendo hook into it, most likely use the same CSS className, https://github.com/RedHatInsights/curiosity-frontend/blob/master/src/components/optinView/optinView.js#L126-L128

### original placement
<img width="1204" alt="Screen Shot 2020-11-09 at 2 04 20 PM" src="https://user-images.githubusercontent.com/3761375/98584952-8b269d80-2294-11eb-93cf-1ad3ba1a5201.png">

cdcabrera:
As of 20201130 this is blocked from migrating forward until the 1.2.0 release is pushed, see #480

cdcabrera:
As of 20201202 there appears GUI work is now ready to proceed, see #521

PR #506, opened 2020-11-16
feat(rhelView, openshiftView): issues/490 add tour button

zanewoodfin:
## What's included

- feat(rhelView, openshiftView): issues/490 add "Take a tour" button

### Notes
A hover effect was not possible on this type of button through Patternfly.  The text of the button has an extra "a" in it from what was shown in the mockup because I reused the text from the previously existing tour button.  Both of these details were approved by UX.

### Local run check
1. `$ yarn start`
1. visit /beta/subscriptions/rhel-sw/all and /beta/subscriptions/openshift-sw
1. confirm presence of tour button on both pages

## Example

take_tour

## Updates issue/story
Updates #490

### DIFF
--- a/src/components/pageLayout/pageHeader.js
+++ b/src/components/pageLayout/pageHeader.js
@@ -4,7 +4,7 @@ import {
   PageHeader as RcsPageHeader,
   PageHeaderTitle
 } from '@redhat-cloud-services/frontend-components/components/cjs/PageHeader';
-import { Button } from '@patternfly/react-core';
+import { Button, Flex, FlexItem, Label as PfLabel } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
 import { helpers } from '../../common';
 import { translate } from '../i18n/i18n';
@@ -14,13 +14,25 @@ import { translate } from '../i18n/i18n';
  *
  * @param {object} props
  * @param {Node} props.children
+ * @param {boolean} props.includeTour
  * @param {string} props.productLabel
  * @param {Function} props.t
  * @returns {Node}
  */
-const PageHeader = ({ children, t, productLabel }) => (
+const PageHeader = ({ children, includeTour, productLabel, t }) => (
   <RcsPageHeader>
-    <PageHeaderTitle title={children} className="pf-u-mb-sm" />
+    <Flex justifyContent={{ sm: 'justifyContentSpaceBetween' }}>
+      <FlexItem>
+        <PageHeaderTitle title={children} className="pf-u-mb-sm" />
+      </FlexItem>
+      <FlexItem>
+        {includeTour && (
+          <Button variant="link" className="uxui-curiosity__button-tour" isInline>
+            <PfLabel color="blue">{t('curiosity-optin.buttonTour')}</PfLabel>
+          </Button>
+        )}
+      </FlexItem>
+    </Flex>
     {productLabel && (
       <p>
         {t(`curiosity-view.subtitle`, { appName: helpers.UI_DISPLAY_NAME, context: productLabel }, [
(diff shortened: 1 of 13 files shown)


src/components/pageLayout/pageHeader.js
```diff
@@ -14,13 +14,26 @@ import { translate } from '../i18n/i18n';
...
+      </SplitItem>
+      <SplitItem isFilled />
+      <SplitItem>
+        {includeTour && (
+          <Button variant="link" className="uxui-curiosity__button-tour">
+            <PfLabel color="blue">{t('curiosity-optin.buttonTour')}</PfLabel>
+          </Button>
+        )}
+      </SplitItem>
+    </Split>
```
cdcabrera, line 36: required/optional: 

First up this totes works as intended!  I feel bad for even asking for this update, especially since the graph is a nightmare on smaller screens.

As you've discovered in PF there are multiple ways to do the same exact thing, kinda... Although we don't really plan for them the `Split` component doesn't appear to stack at smaller sizes, instead it bunches up the title... feels like a prop fix that could be applied to that particular PF component... not sure why they didn't just wrap the `Flex` inside of it anyway other than older browser support maybe... 

> If `Split` is fixed, or there's something we missed, letting it be stacked at smaller screen sizes we have no problem letting this go and sticking with the choice as is!

Lets flip to using the `Flex` and `FlexItem` components here, we already use them on OptIn, so nothing new... they are tricky to use without opening the code up and looking through the props' settings, trying to read through the PF documentation can be frustrating, just go right to the source
- https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/layouts/Flex/Flex.tsx#L257

```
<Flex justifyContent={{ sm: 'justifyContentSpaceBetween' }}>
      <FlexItem>
        <PageHeaderTitle title={children} className="pf-u-mb-sm" />
      </FlexItem>
      <FlexItem>
        {includeTour && (
          <Button variant="link" className="uxui-curiosity__button-tour">
            <PfLabel color="blue">{t('curiosity-optin.buttonTour')}</PfLabel>
          </Button>
        )}
      </FlexItem>
    </Flex>
```

src/components/pageLayout/pageHeader.js
```diff
@@ -56,6 +70,7 @@ PageHeader.propTypes = {
  * @type {{productLabel: null, t: translate}}
  */
 PageHeader.defaultProps = {
+  includeTour: false,
```
cdcabrera, line 73: required:
annotation update

src/components/pageLayout/pageHeader.js
```diff
@@ -46,6 +59,7 @@ const PageHeader = ({ children, t, productLabel }) => (
  */
 PageHeader.propTypes = {
   children: PropTypes.node.isRequired,
+  includeTour: PropTypes.bool,
```
cdcabrera, line 62: required:
annotation update

cdcabrera (changes requested):
Overall fantastic, much appreciated! Especially like making the component optional per-product

Minor things only, everything works, 99.8% of the way there =) 
- 2 minor annotations
- And flipping to the `Flex` component from `Split`

src/components/pageLayout/pageHeader.js
zanewoodfin: Okay, it should be fixed.  I also updated the button to have the isInline prop, so that it wouldn't have button shaped margin around it and would left align all the way.

src/components/pageLayout/pageHeader.js
zanewoodfin: Updated.

cdcabrera (approved):
LGTM!

I went ahead and updated the `dist` check. If that pops up again to update those snapshots just run `$ yarn test:integration-dev` ... it's just a double check for a consistent build output. Every once-in-awhile it updates

For this commit/scenario it's possible because we reused an element/component it may have simply combined some code hence the extra file. Appears consistent after updating my local

ntkathole:
@cdcabrera @zanewoodfin checked on qa-beta env... looks like that button is not working. It won't open tour pop-up on clicking. Am I missing something ?

cdcabrera:
That is kind of correct! The tour button requires syncing with the Pendo team, they'll be handling the full activation, we had hoped using the same ID on it would automagically update it, it did not. Unfortunately the Pendo tooling has no distinction between environments, we're debating on when it should be activated and if it'll affect Prod when it is... We'll keep you in the loop @ntkathole

Issue #492, opened 2020-11-03
Update the per-page minimum


cdcabrera:
## Description
As a developer I need to update the per-page minimum per UX/UI

## Acceptance Criteria

- [x] per-page minimum to 100

## Assumptions and Questions
- review min-height component and how it updates. there may be awkward UX using it when the per-page setting is exceedingly high

PR #507, opened 2020-11-16
fix(openshiftView,rhelView): issues/492 increase per-page results

### LINKED ISSUE #501: Screen adjustments may leave blank space on page
## Description
On some screen adjustments (minimize and maximize), page may leave blank space. I think this is very rare case and minimal priority issue. On page refresh, everything looks good.

**I see this is not new behavior. It is also reproducible on prod as well.**

### Expected behavior
Screen should adjust correctly without refresh. 

#### Steps to reproduce
1. Open subscription watch page (original screen size).
2. Minimize screen.
3. Maximize screen.
4. Apply any filter.
5. See blank page left in between graph and system table.

### Additional context, version info
min-max-screen

cdcabrera:
## What's included

- fix(openshiftView,rhelView): issues/492 increase per-page results
- fix(minHeight): issues/501 restructure resize event 
   * minHeight, restructure resize event, autoUpdate into separate params
- fix(inventoryList): issues/492 set minHeight on first, last page 
   * inventoryList, refresh minHeight for first and last inventory pages
   * styling, reduce inventory label size, aid in paging
- perf(openshiftView,rhelView): simplify displayName callback

### Notes
- Implements a restructure of the minHeight component which helps fix #501 
- Performance update, restructures/simplifies the callbacks used to for the displayName on inventory

### Proxy run check for "per-page" minimum
1. `$ yarn start:proxy`
1. confirm the inventory "per-page" is set to `100` and the corresponding API call reflects `100`

### Proxy run check for #501 
1. `$ yarn start:proxy`
1. mimic the behavior described in #501 . The screen/display should no longer leave a blank area

## Example

...

## Updates issue/story

#501 #492


ntkathole:
VERIFIED
Now per-page limit set to default 100 and also page issue now reproducible (have to clear cache).

Issue #493, opened 2020-11-03
Review passing inventory config for disabling guests display


cdcabrera:
## Description
Review passing inventory config for a callback/function that dictates whether a field should show the guests display (be expandable)

## Acceptance Criteria
- [x] Config for disabling guests' display

## Assumptions and Questions
-

PR #510, opened 2020-11-17
feat(inventoryList): issues/493 product disable guests setting

cdcabrera:
## What's included

- feat(inventoryList): issues/493 product disable guests setting
   * inventoryList, hasGuests callback setting
   * openshiftView, rhelView, apply basic inventoryList settings

### Notes
- behind the scenes update, no visual changes
- applies a settings callback that replaces the `hasGuests` check in `inventoryList.js`

### Proxy run check
1. `$ yarn start:proxy`
1. confirm that guests/VMs **are still expandable** when "guests/VMs are greater than zero" for a listing. there should be no errors present
1. if you wish to confirm the logic works, on one of the product views you can update the `initialInventorySettings` object with the following snippet, where the "greater than one" represents how many guests it takes to disable the expandable section. After updating a view, hit refresh and you should no longer see an expandable section on the targeted row entries.
   ```
      ...
      initialInventorySettings: {
         hasGuests: data => {
            console.log(data);
            const { numberOfGuests = 0, subscriptionManagerId = null } = data;
            return numberOfGuests > 1 && subscriptionManagerId;
         }
     },
      ...
   ```

## Example

...

## Updates issue/story

#493

### DIFF
--- a/src/components/inventoryList/inventoryList.js
+++ b/src/components/inventoryList/inventoryList.js
@@ -99,7 +99,7 @@ class InventoryList extends React.Component {
    * @returns {Node}
    */
   renderTable() {
-    const { filterGuestsData, filterInventoryData, listData, query, session } = this.props;
+    const { filterGuestsData, filterInventoryData, listData, query, session, settings } = this.props;
     let updatedColumnHeaders = [];
 
     const updatedRows = listData.map(({ ...cellData }) => {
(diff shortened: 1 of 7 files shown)


src/components/inventoryList/inventoryList.js
```diff
@@ -113,21 +113,29 @@ class InventoryList extends React.Component {
...
 
-      const hasGuests = cellData?.numberOfGuests > 0;
-      const guestsId = cellData?.subscriptionManagerId;
       updatedColumnHeaders = columnHeaders;
 
+      const guestsId = cellData?.subscriptionManagerId;
+      let hasGuests = cellData?.numberOfGuests > 0 && guestsId;
+
+      // Apply hasGuests callback, return boolean
+      if (typeof settings.hasGuests === 'function') {
```
cdcabrera, line 122: required:
should convert this to `settings?.hasGuests` since it's not a required prop

Issue #512, opened 2020-11-18
cloudmeter mismatch notification state should refresh based on filters 


ntkathole:
## Description
When there is mismatch at SLA/Usage level, mismatch notification should display only for selected filters. Currently if any of the filter is mismatched, notification does not change its state based on selected filter.

### Expected behavior
cloudmeter mismatch notification state should change based on filters.

#### Steps to reproduce
1. Have account with 2 hosts as following :
cloudmeter reported - 2 premium hosts, while,
RHSM reported - 1 premium and 1 standard host
2. Above account will display mismatch when standard and premium sla filter is selected, but it should not show mismatch when unspecified or self-support sla is selected.

### Additional context, version info
As we don't refresh notification based on filters, it may possible that user can see mismatch for some filters, which he/she should not.
![filter-mismatch](https://user-images.githubusercontent.com/16951913/99515415-ef431300-29b2-11eb-8a5d-6138e337ffb1.gif)

cdcabrera:
During initial development we attempted to vary the banner display based on filter as this issue notes. However, the banner kept appearing and disappearing similar to what's displayed in the gif. It was decided via UX @bclarhk @rblackbu that once the banner turned on it should stay on until the end-user either manually closes it, and/or a page refresh happens resetting the banner display parameters. This behavior is described in the initial PR #503 notes area. 

This issue is slightly interrelated to the behavior of #511 fixing this issue may help resolve parts of this behavior, or the behavior we witnessed during initial development. 

Overall this issue highlights a behavior that the banner is working as intended and that NOT turning the banner on and off relates a purposeful UX decision. @bclarhk @rblackbu

bclarhk:
The banner should display if there's a mismatch on _any_ of the RHEL systems. It's positioned above the filters for this reason.

HOWEVER...

It looks like the banner is not displaying on the initial page load. This is wrong but may need a new issue (if it is indeed a bug) to avoid confusion. Is the default result set provided by the API for Tally, Capacity, and Hosts not all inclusive?

cdcabrera:
Breaking out the @bclarhk explanation further... The stated intention for the banner is

>  "[Any system that has a data mismatch means the banner should show outside of the user selected filter behavior]"

[...]

Issue #511, opened 2020-11-18
Mismatch is shown even if has_cloudigrade_mismatch is false


ntkathole:
## Description
"Improve reporting accuracy" notification is displayed even if api returns `has_cloudigrade_mismatch ` as false for today's date.

Env: QA-beta
Account : swatch-rhui

### Expected behavior
"Improve reporting accuracy" notification should display only when `has_cloudigrade_mismatch ` is false.

#### Steps to reproduce
1. Have cloudmeter configured account with no mismatch.
2. Login to account.

### Additional context, version info

```
{'date': datetime.datetime(2020, 11, 18, 1, 2, 51, 327983, tzinfo=tzutc()),
 'instance_count': 1,
 'cores': 0,
 'sockets': 1,
 'physical_instance_count': 0,
 'physical_cores': 0,
 'physical_sockets': 0,
 'hypervisor_instance_count': 0,
 'hypervisor_cores': 0,
 'hypervisor_sockets': 0,
 'cloud_instance_count': 1,
 'cloud_cores': 0,
 'cloud_sockets': 1,
 'has_data': True,
 'has_cloudigrade_data': True,
 'has_cloudigrade_mismatch': False}
```

![Screenshot from 2020-11-18 10-28-42](https://user-images.githubusercontent.com/16951913/99486879-12f26300-298b-11eb-93dc-06d36726203e.png)

ntkathole:
As per conversation with @cdcabrera, current (18/11/2020) logic is, if any one entry in tally response for particular granularity (i.e. last 30 days record for daily) returns false, then we display mismatch data banner. 

We should consider only latest (today's date) entry to display this banner else if any user has mismatch in account for 1 day, he/she may see that banner for upcoming 1 month for daily granularity. This may confuse user as even if account mismatch is fixed, page would display mismatch banner.

ntkathole:
@cdcabrera with above fix, we are considering last 2 days for mismatch, but we missed functionality that show notification when mismatch is at filter (sla/usage) level.
ex: There is no notification in account when /api/rhsm-subscriptions/v1/tally/products/RHEL returns
```
has_cloudigrade_data : true
has_cloudigrade_mismatch : false
```
and /api/rhsm-subscriptions/v1/tally/products/RHEL?sla=premium returns
```
has_cloudigrade_data : true
has_cloudigrade_mismatch : true
```

cdcabrera:
[...]

Issue #499, opened 2020-11-09
Create a current subscriptions daily graph component


cdcabrera:
## Description
As a UX/UI person I need to create a daily display pie chart graph similar to the data currently used in Dashboard

## Acceptance Criteria
- [ ] Screenshot/storyboard
- [ ] Develop a graph card component

## Assumptions and Questions
-

cdcabrera:
As of 20210104 priority for this is low

PR #513, opened 2020-11-18
fix(bannerMessages): issues/511 limit activation to current date

cdcabrera:
## What's included

- perf(rhsmServices,userServices): expose cancel options
- fix(bannerMessages): issues/511 limit activation to current date
   * dateHelpers, expose a currentDateTime range
   * bannerMessages, custom API call for reports/tally
   * rhsmActions, restructure for cancel id, add getMessageReports
   * messagesReducer, apply getMessageReports action type to state
   * appMessagesSelectors, sync state with messagesReducer
   * rhsmTypes, clean up, new message reports type

### Notes
- the logic emulates the daily logic used on the platform Dashboard, that pulls 2 days worth of Tally entries

### Proxy run check
1. `$ yarn start:proxy`
1. log in with an account with Cloud Meter data and the `is mismatched` flag is `true` within the last 2 days
1. confirm the banner shows
1. double check that older Cloud Meter data and the `is mismatched` flag is `true` older than the last 2 days does not fire the banner

## Example

...

## Updates issue/story

Relates #499 #512 #502 
Updates #511


cdcabrera:
Should be good to go, just remember to do the "Rebase and merge"

PR #518, opened 2020-11-30
fix(bannerMessages): issues/511 apply query filter 

cdcabrera:
## What's included

- fix(bannerMessages): issues/511 copy update
- fix(bannerMessages): issues/511 apply query filter 
   * bannerMessages, query updates for banner display
   * openshiftView, rhelView, pass query to banner
   * appMessageSelectors, query pass-through

### Notes
- Applies the query filter to the banner. This in turn adds an additional API call that fires on filter update.
   - The initial request was around the SLA filter... this should apply all graph related filters inclusive of UOM, and the exception of NO user applied granularity is set towards the banner message.
- Minor copy adjustment per UX
- @ntkathole 

### Proxy run check
1. `$ yarn start:proxy`
1. log in with the appropriate account that has mismatched Cloud Meter data within the last 2 days based on the appropriate SLA or Usage filter selection
1. with the banner message not initially displaying... confirm that when selecting the SLA or Usage filters the banner message now appears 

#### Copy update
Screen Shot 2020-11-30 at 3 50 22 PM

## Updates issue/story

#511

### DIFF
--- a/src/components/bannerMessages/bannerMessages.js
+++ b/src/components/bannerMessages/bannerMessages.js
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Alert, AlertActionCloseButton, AlertVariant, Button } from '@patternfly/react-core';
 import { ExternalLinkAltIcon } from '@patternfly/react-icons';
-import { connect, reduxActions, reduxSelectors } from '../../redux';
+import _isEqual from 'lodash/isEqual';
+import { apiQueries, connect, reduxActions, reduxSelectors } from '../../redux';
 import { translate } from '../i18n/i18n';
 import { dateHelpers, helpers } from '../../common';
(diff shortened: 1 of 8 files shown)


src/components/bannerMessages/bannerMessages.js
```diff
@@ -125,6 +129,7 @@ BannerMessages.propTypes = {
  */
 BannerMessages.defaultProps = {
   getMessageReports: helpers.noop,
+  query: {},
```
cdcabrera, line 132: required:
apply annotations

src/components/bannerMessages/bannerMessages.js
```diff
@@ -107,6 +110,7 @@ class BannerMessages extends React.Component {
 BannerMessages.propTypes = {
   appMessages: PropTypes.object.isRequired,
   getMessageReports: PropTypes.func,
+  query: PropTypes.object,
```
cdcabrera, line 113: required:
apply annotations

Issue #539, opened 2020-12-15
Activate Satellite for left nav, display fixes


cdcabrera:
## Description
As a developer I need to update the Platform left menu config to activate the left navigation for ~Satellite 6~ Satellite and Satellite Capsule and Server ~Satellite 6 Capsule~.

## Acceptance Criteria
- [x] Platform config updated
   - initial CI-beta PR
   - QA, and related prod PRs
- [x] Strings for Sat Capsule & Server updated #540 #542  #553
- [x] "Clear" state behavior fixed #541 

## Assumptions and Questions
- relates #524
- there are currently strings around the title for Capsule that need to be updated, they are lengthy. we may be able to hold and wait for further updates

ntkathole:
Adding just for a record, this needs api fixes https://issues.redhat.com/browse/ENT-3447 and https://issues.redhat.com/browse/ENT-3363

cdcabrera:
resolved on 20220914 https://github.com/RedHatInsights/cloud-services-config/pull/1287

Issue #524, opened 2020-12-07
Activate Satellite display


cdcabrera:
## Description
As a developer I need to activate the Satellite 6 display

## Acceptance Criteria
- Platform left navigation yml updated, see #539
- [x] Satellite components activated
   - product view

## Assumptions and Questions
- Satellite 6, Satellite 6 Capsule, see https://raw.githubusercontent.com/RedHatInsights/rhsm-subscriptions/master/src/main/resources/product_id_to_products_map.yaml
- Applying a 3rd product should help us refine the architecture for "generic/basic" product displays. This would help with potential future base product displays that can be setup with even less effort.
   - Even if the majority of future products are custom there would still need to be a base for state management and possible key component displays

PR #540, opened 2020-12-15
fix(productView): issues/539 product label string

cdcabrera:
## What's included

- fix(productView): issues/539 product label string
   * productView, reintroduce productLabel, props
   * productViewSatellite, props update,
   * routerTypes, viewParameter

### Notes
- Sat 6 Capsule page title now displays Sat 6

### Proxy run check
1. `$ yarn start:proxy`
1. navigate to the following routes
   - `/beta/subscriptions/satellite-sw/all`
   - `/beta/subscriptions/satellite-sw/capsule`
1. confirm no browser errors are present, and the the related title and the graph tooltip and legend copy accurately relates "Satellite"

## Example

...

## Updates issue/story

#542 #524 #539


src/components/productView/__tests__/productView.test.js
```diff
@@ -8,7 +8,7 @@ describe('ProductView Component', () => {
       productConfig: {},
       routeDetail: {
         pathParameter: 'lorem ipsum',
-        productParameter: 'dolor sit'
+        viewParameter: 'dolor sit'
```
cdcabrera, line 11: required:
be helpful to include the prop that produces the productLabel since the fix for this revolves around it

src/components/productView/__tests__/productView.test.js
```diff
@@ -21,7 +21,7 @@ describe('ProductView Component', () => {
       productConfig: {},
       routeDetail: {
         pathParameter: null,
-        productParameter: null
+        viewParameter: null
```
cdcabrera, line 24: required:
include prop that produces productLabel

Issue #546, opened 2021-01-11
system table page navigation is broken


ntkathole:
## Description
System table page navigation is broken. Could not naviagte to previous page as back button is disabled and numbers not shown currently in table. (see screenshot attached)

### Expected behavior
Correct page navigation and system table numbers. 

#### Steps to reproduce
1. Reproducible in CI-beta only.
2. Navigate to https://ci.cloud.redhat.com/beta/subscriptions/rhel-sw/all
3. Navigate to 2nd page and try to click on previous page button to navigate back.
4. Button is disabled.

### Additional context, version info
![Screenshot from 2021-01-11 20-34-44](https://user-images.githubusercontent.com/16951913/104199303-48787180-544d-11eb-8259-69f62d983483.png)

cdcabrera:
As of 20200111 on our local proxy run we rolled back recent changes all the way to #538 the last of the Dec updates and the issue still persists. At the moment this coming across as not on the Curiosity side, but something else possibly platform or api, we'll keep looking

cdcabrera:
@ntkathole this should be ready for confirmation now

ntkathole:
Fixed! Thanks :+1:

PR #522, opened 2020-12-02
feat(inventorySubscriptions): issues/491 basic inventory

cdcabrera:
## What's included

- feat(inventorySubscriptions): issues/491 basic inventory
   * locale, tab and subscriptions strings
   * guestsList, query update
   * bannerMessages, query update
   * inventoryList, query, title adjust
   * inventorySubscriptions, new component
   * openshiftView,rhelView, apply inventory tabs, sort filters
   * pagination, restructure for multiple inventories
   * viewReducer, restructure for multiple inventory queries
   * inventoryListSel, subscriptionsListSel, query check
   * queryTypes, sort, dir types for multiple inventories
   * styling, card title adjustment
- fix(inventorySubscriptions): issues/491 deactivate inventory

### Notes
- Apply and then deactivate subscriptions inventory display @bclarhk @rblackbu 
- This is the base set of components and state behavior, obvious adjustments will need to be made for end-result column display
- Merging this will cause minimal visual display changes. We've deactivated the subscriptions inventory display while waiting on the API, see below screenshot for below visual updates

### Deactivated display
Screen Shot 2020-12-11 at 12 04 41 AM

### Activated display
Screen Shot 2020-12-11 at 12 06 29 AM

## Updates issue/story

#491

PR #547, opened 2021-01-11
fix(pagination): issues/491 missing limit, offset query

cdcabrera:
## What's included

- fix(pagination): issues/491 missing limit, offset query
   * pagination, pass missing query from inventory
   * inventoryList, inventorySubs, pass query to pagination
- perf(pagination): move redux on page into inventory * pagination, restructure, remove redux
   * paginationHelpers, page to offset, offset to page
   * inventoryList, inventorySubs, apply onPage events

### Notes
- There was an oversight on passing the specific/unique queries for limit and offset from the parent inventory components. 
- ~The query parameter is now required for pagination, in the future code checks should catch the query if it fails to pass.~ To avoid "prop-drilling", and overly-complicated type props used for updating Redux state, we now handle paging in the inventory components, this avoids the entire original issue. Unit tests have been updated accordingly.
   - We're leaving the original fix commit for now, we may squash it on merge
- It appears this has been broken since #522 this was further compounded with a minimally functional host inventory display in the CI environment. Unit tests may still need to be expanded

### Proxy run check
1. `$ yarn start:proxy`
1. login with an account that multiple pages of inventory entries
1. page to the next page, and then a subsequent page to confirm inventory updates

## Example

...

## Updates issue/story

#546

### DIFF
--- a/src/components/inventoryList/inventoryList.js
+++ b/src/components/inventoryList/inventoryList.js
@@ -206,6 +206,7 @@ class InventoryList extends React.Component {
                 perPageDefault={updatedPerPage}
                 offsetType={reduxTypes.query.SET_QUERY_RHSM_HOSTS_INVENTORY_TYPES[RHSM_API_QUERY_TYPES.OFFSET]}
                 limitType={reduxTypes.query.SET_QUERY_RHSM_HOSTS_INVENTORY_TYPES[RHSM_API_QUERY_TYPES.LIMIT]}
+                query={query}
               />
             </CardActions>
           </CardHeader>
@@ -242,6 +243,7 @@ class InventoryList extends React.Component {
                 dropDirection="up"
                 offsetType={reduxTypes.query.SET_QUERY_RHSM_HOSTS_INVENTORY_TYPES[RHSM_API_QUERY_TYPES.OFFSET]}
                 limitType={reduxTypes.query.SET_QUERY_RHSM_HOSTS_INVENTORY_TYPES[RHSM_API_QUERY_TYPES.LIMIT]}
+                query={query}
               />
             </TableToolbar>
           </CardFooter>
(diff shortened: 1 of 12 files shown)


src/components/inventorySubscriptions/inventorySubscriptions.js
```diff
@@ -162,9 +187,13 @@ class InventorySubscriptions extends React.Component {
       );
     }
 
-    const updatedPerPage = query?.[RHSM_API_QUERY_TYPES.LIMIT] || perPageDefault;
-    const updatedPage = query[RHSM_API_QUERY_TYPES.OFFSET] / query[RHSM_API_QUERY_TYPES.LIMIT] + 1 || 1;
-    const isLastPage = updatedPage === Math.ceil(itemCount / updatedPerPage);
+    const updatedPerPage = query[RHSM_API_QUERY_TYPES.LIMIT] || perPageDefault;
+    const updatedOffset = query[RHSM_API_QUERY_TYPES.OFFSET];
+    const isLastPage =
```
cdcabrera, line 192: optional: since we're already offsetting basic paging calcs in helpers, maybe move "isLastPage" into the paginationHelpers too

src/components/inventoryList/inventoryList.js
```diff
@@ -182,9 +207,13 @@ class InventoryList extends React.Component {
       );
     }
 
-    const updatedPerPage = query?.[RHSM_API_QUERY_TYPES.LIMIT] || perPageDefault;
-    const updatedPage = query[RHSM_API_QUERY_TYPES.OFFSET] / query[RHSM_API_QUERY_TYPES.LIMIT] + 1 || 1;
-    const isLastPage = updatedPage === Math.ceil(itemCount / updatedPerPage);
+    const updatedPerPage = query[RHSM_API_QUERY_TYPES.LIMIT] || perPageDefault;
+    const updatedOffset = query[RHSM_API_QUERY_TYPES.OFFSET];
+    const isLastPage =
```
cdcabrera, line 212: optional: since we're already offsetting basic paging calcs in helpers, maybe move "isLastPage" into the paginationHelpers too

Issue #476, opened 2020-10-26
Filter inventory by Display Name


cdcabrera:
## Description

As a UI/UX person I need to consider large numbers of inventory entries and the user ability to filter by a string associated with the inventory Display Name.

## Acceptance Criteria
- [ ] Storyboard or screenshot confirmation
- [ ] Toolbar component for tabs, relates towards #407 
   - Inventory level toolbar with display name filter
   

## Assumptions and Questions

- Consider creating a "widget" a self contained form field capable of updating state standalone
- Should this be a separate toolbar specifically for inventory, 3 toolbars?
- Should this apply to the global app search and apply to the Tally (graph display) as well or be limited to inventory display only
- Dev story #475

rblackbu:
![Uploading Screen Shot 2020-11-05 at 11.50.24 AM.png…]()

rblackbu:
<img width="1230" alt="Screen Shot 2020-11-05 at 12 02 19 PM" src="https://user-images.githubusercontent.com/25234746/98272572-3ebf2300-1f5f-11eb-9bb0-ff2a5169409f.png">

rblackbu:
<img width="1308" alt="Screen Shot 2020-11-05 at 11 50 24 AM" src="https://user-images.githubusercontent.com/25234746/98272704-60200f00-1f5f-11eb-9894-ce4640a6bc3c.png">

cdcabrera:
Intended behavior, confirmation
First round
- the filter should always display
- should only be the input field with magnifying glass
- still need input on "clearing" the field

API wise
- looking at doing an "autocomplete" with 2-3 characters. We should ask API team for input

PF wise
- still looking for the Patternfly autocomplete example

PR #556, opened 2021-01-21
fix(inventoryList): issues/476 show field on zero results

cdcabrera:
## What's included

- fix(inventoryList): issues/476 show field on zero results
- fix(toolbarFieldDisplayName): issues/476 allow enter submit

### Notes
- based on UX/design feedback reactivates the display name field. accounts for scenarios in which zero results normally hides "paging". the new behavior is to now display paging and the display name field when there are "zero" results
- enables a user to use the enter/return key to submit the display name field
- have a question around the paging disabled state, do we want to remove the grey disabled color? @bclarhk @rblackbu 

### Proxy run check
1. `$ yarn start:proxy`
1. login with an account having a hosts inventory
1. navigate to a RHEL architecture and enter a display name search that yields zero results, the result should be similar to the below screenshot example
   - Also confirm that switching between RHEL architectures both maintains the value in the display name field, adjusts the results based off of "all", and allows you to still edit display name

### Accepted solution
Screen Shot 2021-01-21 at 11 01 08 AM

### NOT accepted, prior solution with paging disabled and showing
Screen Shot 2021-01-21 at 10 13 29 AM

## Updates issue/story

#476

### DIFF
--- a/src/components/form/__tests__/__snapshots__/checkbox.test.js.snap
+++ b/src/components/form/__tests__/__snapshots__/checkbox.test.js.snap
@@ -109,6 +109,7 @@ Object {
   "checked": true,
   "currentTarget": Object {},
   "id": "generatedid-",
+  "keyCode": undefined,
   "name": "generatedid-",
   "persist": [Function],
   "target": Object {},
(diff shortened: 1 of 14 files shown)


bclarhk (approved):
Nice and clean. I approve :-D

PR #576, opened 2021-02-04
fix(build): npm updates

cdcabrera:
## What's included

- fix(build): npm updates
- fix(build): pf, platform npm updates

### Notes
- There were build fails upon merging the original npm updates PR #575 so they were rolled out. This is a more focused set of updates aiming to eliminate the build fails until we can figure out the specific package causing issues.

1. `$ yarn test`

### Check the build
1. `$ yarn build`
1. no errors should be present

## Example

...

## Updates issue/story

Ongoing


cdcabrera:
The failure was corrected when we cleared Travis' build cache for this branch. It's unclear why that wasn't performed automagically as it has in the past.

PR #588, opened 2021-02-22
WIP perf(productViewOpenShift): openshiftView to productViewOpenShift

cdcabrera:
## What's included

- perf(productViewOpenShift): openshiftView to productViewOpenShift

### Notes
- moves openshift to the same product view component that RHEL and Satellite use
- applies a product context that can be consumed through hooks
- similar product tests performed on RHEL and Satellite should be performed on OpenShift
- This update allows us the ability to simplify our component structures enabling more efficient reuse based on "context" instead of prop drilling

1. `$ yarn test`
1. confirm tests come back without error

## Example

...

## Updates issue/story

ent-3324


cdcabrera:
As of 20210302 this PR contains parallel work that has merged with #591 , we can still utilize the branch for artifact purposes in the short term

PR #604, opened 2021-03-18
fix(viewReducer): ent-3644 apply productId to inventory reset

cdcabrera:
## What's included

- fix(viewReducer): ent-3644 apply productId to inventory reset
   * inventoryList, productId to displayName
   * productView, OpenShiftContainer,Dedicated, productId to granularity
   * toolbar, productId to inventory reset
   * toolbarFieldDisplayName, granularity, ranged, uom, apply productId
   * viewReducer, restructure reset to apply specific productId

### Notes
- It appears we discovered a flaw/bug in how we reset the inventory displays. Originally we didn't have dual column displays so we would generically reset the inventory per view, this encompassed all productIds under a viewId. The dual column display created a scenario where we can't/shouldn't generically reset all inventory states. Now **when an inventory reset needs to happen it can no longer be generic, it has to be specific**. 
- There are two solutions we've reviewed
    - **THIS PR** Go with what we know works and apply prop drilling for the `productId` alongside the `viewId`, or
    - #606 Apply a unique solution by retaining the `viewId` and creating a dictionary for each viewId that relates all possible `productIds` contained under a `viewId` enabling us to reset state on those specific productIds

### Proxy run check
1. `$ yarn start:proxy`
1. confirm the inventory paging/offset is reset to the initial offset, in this case 0.
    - all products and their facets should be confirmed

## Example

...

## Updates issue/story

ent-3644

### DIFF
--- a/src/components/inventoryList/__tests__/__snapshots__/inventoryList.test.js.snap
+++ b/src/components/inventoryList/__tests__/__snapshots__/inventoryList.test.js.snap
@@ -231,6 +231,7 @@ exports[`InventoryList Component should handle variations in data: filtered data
     >
       <CardHeaderMain>
         <ToolbarFieldDisplayName
+          productId="lorem"
           t={[Function]}
           value={null}
           viewId="inventoryList"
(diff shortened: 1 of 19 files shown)


cdcabrera:
We're opting for using #606 . The reasoning behind it being
- minimal alteration to existing components, maintains a minimal level of knowledge needed to apply updates
- future components that need to be reset in batches now have a model
- parallels the previous user experience closer than #604 

The only downside we could come up with
- it introduces yet another way to use the routing and nav configs

PR #605, opened 2021-03-18
fix(routerHelpers): ent-3541 rhcloud-12690 alt for search, hash

cdcabrera:
## What's included

- fix(routerHelpers): ent-3541 rhcloud-12690 alt for search, hash
   * build, update proxy, spandx config to account for left nav changes
   * routerHelpers, alt for location.href, use search and hash directly

### Notes
- Related to https://github.com/RedHatInsights/cloud-services-config/pull/456
- Corrects the dependency on window.location.href and instead directly uses `hash` and `search`. There is a race like condition that happens where sometimes window.location.href is updated before the route shift, sometimes it isn't
- This update affects all product views, OpenShift and the future views for Satellite

### Proxy run check
1. `$ yarn start:proxy`
1. Navigate to `/beta/insights/subscriptions/rhel-arm` or any RHEL facet other than `/rhel`
1. Now attempt to navigate towards `/beta/insights/subscriptions/rhel` using the left nav ... RHEL should now appear. Prior to this update RHEL would not display.

## Example

...

## Updates issue/story

ent-3541

### DIFF
--- a/src/components/router/__tests__/routerHelpers.test.js
+++ b/src/components/router/__tests__/routerHelpers.test.js
@@ -8,10 +8,11 @@ import {
 } from '../routerHelpers';
 
 describe('RouterHelpers', () => {
-  const mockWindowLocationProp = async ({ prop, url, callback }) => {
+  const mockWindowLocation = async ({ url, callback }) => {
+    const updatedUrl = new URL(url);
     const { location } = window;
     delete window.location;
-    window.location = { [prop]: url };
+    window.location = { href: updatedUrl.href, search: updatedUrl.search, hash: updatedUrl.hash };
     await callback();
     window.location = location;
   };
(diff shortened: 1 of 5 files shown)


src/components/router/__tests__/routerHelpers.test.js
```diff
@@ -146,7 +147,7 @@ describe('RouterHelpers', () => {
       }
     });
 
-    mockWindowLocationProp({
+    mockWindowLocation({
       prop: 'href',
```
cdcabrera, line 151: required: remove `prop: 'href',` lines 151, 161, 171

src/components/router/routerHelpers.js
```diff
@@ -55,9 +55,18 @@ const getNavigationDetail = ({ id = null, pathname = null, returnDefault = false
...
+   * Originally, we used a split on window.location.href.split(navigationItem.path) as an easy way
+   * to pull everything and maintain sequence, but there appears to be a race like condition
+   * happening on certain routes where window.location.href is updated first in some routing
+   * instances, and then in other instances requires being updated by the app/GUI.
+   */
   if (navigationItem) {
-    const [, routeHref] = window.location.href.split(navigationItem.path);
-    navigationItem.routeHref = `${navigationItem.path}${routeHref || ''}`;
+    const { search = '' } = window.location;
+    const { hash = '' } = window.location;
```
cdcabrera, line 67: required: these should be combined `const { search = '', hash = '' ...`

PR #614, opened 2021-04-09
fix(toolbarFieldRangedMonthly): ent-3748 apply position right

cdcabrera:
## What's included

- fix(toolbarFieldRangedMonthly): ent-3748 apply position right
   * select, expand display options, allow up down and left right
   * toolbarFieldRangedMonthly, apply position right
   * style, position right for select components

### Notes
- PF Select doesn't appear to have a position left/right. We repurposed the CSS from PF Dropdown and it appears to work!

### Proxy run check
1. `$ yarn start:proxy`
1. open OpenShift Container or OpenShift Dedicated and confirm the monthly ranged dropdown now has its position flush right

## Example

Screen Shot 2021-04-09 at 11 02 33 AM

Screen Shot 2021-04-09 at 11 01 45 AM

## Updates issue/story

ent-3748

### DIFF
--- a/src/components/form/select.js
+++ b/src/components/form/select.js
@@ -1,11 +1,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Select as PfSelect, SelectOption as PfSelectOption, SelectVariant } from '@patternfly/react-core';
+import {
+  DropdownDirection,
+  DropdownPosition,
+  Select as PfSelect,
+  SelectOption as PfSelectOption,
+  SelectVariant
+} from '@patternfly/react-core';
 import _cloneDeep from 'lodash/cloneDeep';
 import _isEqual from 'lodash/isEqual';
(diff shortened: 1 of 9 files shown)


src/components/form/select.js
```diff
@@ -247,6 +281,7 @@ class Select extends React.Component {
 Select.propTypes = {
   ariaLabel: PropTypes.string,
   className: PropTypes.string,
+  direction: PropTypes.oneOf(Object.values(SelectDirection)),
```
cdcabrera, line 284: required: update jsdoc annotation

src/components/form/select.js
```diff
@@ -290,6 +326,7 @@ Select.propTypes = {
 Select.defaultProps = {
   ariaLabel: 'Select option',
   className: '',
+  direction: SelectDirection.down,
```
cdcabrera, line 329: required: update jsdoc annotation

PR #656, opened 2021-05-04
refactor(config): ent-2745 consolidate nav, routing configs

cdcabrera:
## What's included

- refactor(config): ent-2745 consolidate nav, routing configs
   * routerConfig, move const to routerHelpers, clean up comps and redux refs
   * routerHelpers, generate nav, route, product groups, add routerConfig const
- refactor(routerHelpers): ent-2745 consolidate nav, route lookup
   * productViewMissing, switch to route config, allow alias matching
   * router, switch to combined nav route detail lookup
   * routerHelpers, consolidate nav, route lookup, clean up
- refactor(productViewMissing): ent-2745 use route helper
- refactor(router): ent-2745 convert class component

### Notes
- restructures the navigation, routing configs into a more cohesive structure.
- intent is to make updating/adding new product routes easier without having to update multiple points
- there's still a minor "todo" to clean up the independent route and nav details used to provide context through routing. this could also be handled when we implement product context

1. `$ yarn test`
1. confirm tests pass

### Proxy run check
1. `$ yarn start:proxy`
1. login and confirm all products and their respective facets are still accessible

## Example

...

## Updates issue/story

ongoing
ent-2745


src/components/router/__tests__/__snapshots__/routerHelpers.test.js.snap
```diff
@@ -283,3 +335,435 @@ Object {
...
+]
+`;
+
+exports[`RouterHelpers should return specific properties: platformLandingRedirect 1`] = `"/"`;
+
+exports[`RouterHelpers should return specific properties: platformModalRedirect 1`] = `"/?not_entitled=subscriptions"`;
+
+exports[`RouterHelpers should return specific properties: productGroups 1`] = `
+Object {
+  "null": Array [
```
cdcabrera, line 569: required: need to ignore `null` entries when generating productGroups

src/components/router/routerHelpers.js
```diff
@@ -29,6 +51,101 @@ const baseName = (helpers.TEST_MODE && '/') || (helpers.DEV_MODE && '/') || dyna
...
+ * Generate product groups for applying query filter resets.
+ *
+ * @param {Array} config
+ * @returns {Array}
+ */
+const generateProductGroups = (config = routeConfig) => {
+  const productGroups = {};
+
+  config.forEach(({ pathParameter, productParameter }) => {
+    const viewIds = ((Array.isArray(productParameter) && productParameter) || [productParameter]).map(
```
cdcabrera, line 92: required: look at filtering `null`

src/components/router/routerHelpers.js
```diff
@@ -43,20 +160,20 @@ const getErrorRoute = routes.find(route => route.activateOnError === true) || {}
  * @param {string} params.id
  * @param {string} params.pathname
  * @param {boolean} params.returnDefault
+ * @param {Array} params.nav
  * @returns {object}
  */
-const getNavigationDetail = ({ id = null, pathname = null, returnDefault = false }) => {
-  const defaultItem = returnDefault && navigation.find(item => item.default === true);
+const getNavigationDetail = ({ id = null, pathname = null, returnDefault = false, nav = navigation }) => {
```
cdcabrera, line 166: optional: could use `navigation: nav = navigation`

src/components/router/routerHelpers.js
```diff
@@ -85,18 +202,19 @@ const getNavigationDetail = ({ id = null, pathname = null, returnDefault = false
  * @param {object} params
  * @param {string} params.id
  * @param {string} params.pathname
+ * @param {Array} params.rt
  * @returns {object}
  */
-const getRouteDetail = ({ id = null, pathname = null }) => {
+const getRouteDetail = ({ id = null, pathname = null, rt = routes }) => {
```
cdcabrera, line 208: optional: could use `routes: rt = routes`

cdcabrera (commented):
required:
need an overarching test/check for the configs

src/components/router/routerHelpers.js
```diff
@@ -29,6 +51,105 @@ const baseName = (helpers.TEST_MODE && '/') || (helpers.DEV_MODE && '/') || dyna
...
+ * Review removing "navigation" and using the route config directly.
+ */
+/**
+ * Generate navigation list, used to help in look ups for "missing products" and provide additional route details.
+ *
+ * @param {Array} config
+ * @returns {{path: string, default: boolean, component: Node, productParameter: string,
+ *     viewParameter: string, activateOnError: boolean, id: string, isSearchable: boolean, pathParameter: string}[]}
+ */
+const generateNavigation = (config = routesConfig) =>
```
cdcabrera, line 65: required:
lets go ahead and try to remove the necessity for "generateNavigation" entirely. as a subsequent commit restructure the helpers associated with pulling product and routing information

src/components/router/routerHelpers.js
```diff
@@ -29,6 +51,100 @@ const baseName = (helpers.TEST_MODE && '/') || (helpers.DEV_MODE && '/') || dyna
...
+/**
+ * Generate routes to be consumed by router.
+ *
+ * @param {Array} config
+ * @returns {Array}
+ */
+const generateRoutes = (config = routesConfig) => {
+  const updatedRoutes = [];
+
+  config.forEach(({ activateOnError, component, disabled, path: routePath, redirect }) =>
```
cdcabrera, line 126: required: now that the condition is removed we can turn this into a direct `=> config.map`

src/components/router/router.js
```diff
@@ -40,7 +39,7 @@ class Router extends React.Component {
               path={item.to}
               strict={item.strict}
               render={({ location, ...routeProps }) => {
-                const navDetail = routerHelpers.getNavigationDetail({
+                const navDetail = routerHelpers.getNavRouteDetail({
```
cdcabrera, line 42: review: is this necessary anymore? this made sense when nav and route were separated but it appears we can just dump the route config props in at the same time instead of calling them after-the-fact. there is a minor mapping for the product and view params to consider

src/components/router/routerHelpers.js
```diff
@@ -37,30 +118,37 @@ const basePath =
...
  */
-const getNavigationDetail = ({ id = null, pathname = null, returnDefault = false }) => {
-  const defaultItem = returnDefault && navigation.find(item => item.default === true);
-  let navigationItem;
+const getNavRouteDetail = ({ id = null, pathname = null, returnDefault = false, config = routesConfig }) => {
+  let navRouteItem;
 
   if (id) {
-    navigationItem = navigation.find(item => item.id === id);
+    navRouteItem = config.find(item => item.id === id);
```
cdcabrera, line 136: review: in looking at how we call this ID lookup, as part of the platform nav interaction, we can simplfy this into its own functionality, or remove the other functionality contained in `getNavRouteDetail` if we consider dumping these props directly into the router using `generateRoutes`

cdcabrera: result: we'll maintain this functionality in the short term, and review again once we start implementing product context w/ hooks and the `react-router-dom` hook `useRouteMatch`

src/components/router/routerHelpers.js
```diff
@@ -29,6 +51,84 @@ const baseName = (helpers.TEST_MODE && '/') || (helpers.DEV_MODE && '/') || dyna
...
+      pathParameter,
+      productParameter,
+      redirect,
+      ...routeConfig
+    }) => ({
+      activateOnError,
+      component,
+      disabled,
+      exact: true,
+      routeConfig: {
```
cdcabrera, line 111: optional: consider renaming this to "productConfig" its primary purpose is to provide identifiers to products

src/components/router/router.js
cdcabrera: maintaining with a minor rename for now, review once context and hooks start taking over. review using `react-router-dom`s `useRouteMatch`

src/components/router/routerHelpers.js
cdcabrera: slightly different direction

src/components/productView/productViewMissing.js
```diff
@@ -27,41 +25,21 @@ const ProductViewMissing = ({ basePath, products, t }) => {
...
    *
    * @event onUpdateHistory
-   * @param {string} path
+   * @param {string} id
    * @returns {void}
    */
-  const onClick = path => history.push(path);
+  const onClick = id => {
+    const { routeHref } = routerHelpers.getRouteConfig({ id });
+    history.push(routeHref);
```
cdcabrera, line 41: future: being able to minimally confirm the onclick is working would help for unit testing. 

notes:
appears that emulating `react-router-dom`s `history` comes back with the appearance of some false positives, the click event "doesn't not say" it fired, tests pass, but doesn't actually push the passed `routeHref`, little deceptive. An alternative we explored... works every time by passing `useHistory` as a prop and overriding it during testing. we do something similar with translate

PR #679, opened 2021-05-20
fix(routes,router): ent-2745 dynamic lazy load views 

cdcabrera:
## What's included

- fix(routes,router): ent-2745 dynamic lazy load views 
   * routes, component views to file paths
   * router, restructure for lazy load of views

### Notes
- generates lazy loaded product views from config "in router" instead of referencing lazy loaded views "in config"
   - a side effect of generating the views is a further breakout in how the gui is code split 

### Proxy run check
1. `$ yarn start:proxy`
1. login and confirm all product views are still able to be accessed

## Example

...

## Updates issue/story

relates ent-2745
ongoing


tests/__snapshots__/dist.test.js.snap
```diff
@@ -7,37 +7,253 @@ Array [
...
+  "./build/static/js/91*chunk*map",
+  "./build/static/js/91*chunk.js",
+  "./build/static/js/92*chunk*map",
+  "./build/static/js/92*chunk.js",
+  "./build/static/js/93*chunk*map",
+  "./build/static/js/93*chunk.js",
+  "./build/static/js/94*chunk*map",
+  "./build/static/js/94*chunk.js",
+  "./build/static/js/95*chunk*map",
+  "./build/static/js/95*chunk.js",
```
cdcabrera, line 250: required
looks like we solved not including snapshots, but this still appears to be including unit tests?

cdcabrera (commented):
need to fix build output

tests/__snapshots__/dist.test.js.snap
cdcabrera: wp exclude corrected it

PR #689, opened 2021-05-26
refactor(routes,productViews): ent-2745 apply product configs to routing

cdcabrera:
## What's included

- refactor(routes,productViews): ent-2745 apply product configs to routing
   * config, add products, update routing with products
   * productView, adjust for using array of product configs
   * productViewDefault, new default product component
   * productViewRhel, OpenShift, Satellite, default props to configs
- fix(select): squash popper test warnings

#### Product configs
- There should be no visual changes associated with this refactor
- **This is an incremental refactor** designed to prep for...
   - applying product context to underlying components
   - generating the route config from product configs
      -  There are unused properties under the product configs around "product id" and "view id" these are currently set in routerHelpers.js. The intent is to transition to these properties

#### Test warnings on select components
Appears there is a local test run warning/error that gets generated from PF's use of the plugin Popper. This warning does not appear in our GitHub actions or Travis run tests.

Our tests worked normal prior to #740 . After adjustments to how we test the select component we now receive warnings around testing Popper, even though we do not directly use that component. This could be due to a mock not being cleared, or related to tests that expand the select, or some other reason. The behavior is mostly odd because the components that are being flagged we focus on related Redux state behavior and the result snapshots are still valid.

We're temporarily squashing this "warning" since it has the appearance of a false positive. We can revisit removing this check at a later date

1. `$ yarn test`
1. confirm tests complete successfully

### Local run check
1. `$ yarn start`
1. confirm app can be navigated and viewed successfully

### Proxy run check
1. `$ yarn start:proxy`
1. confirm app can be navigated and viewed successfully

### Check the build
1. `$ yarn build`
1. confirm build completes successfully

## Example

...

## Updates issue/story

ent-2745


cdcabrera (commented):
snapshot for routerHelpers is rather large, focus the tests, reduce the snapshot footprint

PR #695, opened 2021-06-02
fix(productViewMissing): ent-3913 apply platform appNavClick

cdcabrera:
## What's included

- fix(productViewMissing): ent-3913 apply platform appNavClick
   * testing, mock appNavClick, hook
   * productViewMissing, useRouter, setAppNav
   * platformActions, expose setAppNav
   * platformServices, expose appNavClick, replace navigation
   * useRouter, pass useHistory, proxy useHistory push, setAppNav
- fix(redirect): ent-3833 restructure, force a path 
   * testing, window.replace noop
   * authentication, remove isRedirect
   * productViewMissing, on available product cards redirect
   * redirect, restructure, force a path, pass search, hash
   * router, var rename, centralize exports, index.js
   * routerHelpers, optional chain

### Notes
- Each of these commits covers a different form of redirect
   - **fix redirect** covers handling a forced redirect with `window.location.href` 
   - **fix productViewMissing** covers using the platforms `appNavClick` chrome method (allows updating the left hand nav without forcing a refresh)

### Proxy run check
 since the platform still appears to have some redirects in place, and this automagically performs the redirect you can test 2 ways to confirm it's actually doing something.

#### First
1. `$ yarn start:proxy`
1. confirm the redirect happens by purposefully using a botched url, i.e. `/beta/subscriptions/` and confirm that the product id `rhel` is used and the end result is `/beta/subscriptions/rhel`

#### Second
1. `$ yarn start:proxy`
1. navigate to a purposefully broken path `/beta/subscriptions/`
1. modify the `./src/components/productViewMissing.js` file default prop `availableProductsRedirect` to zero, hit refresh and the card view of missing products should appear
1. click on a card and the path should update accordingly

* _NOTE: testing the left menu highlighting may produce a false positive if the platform redirects are still in place_

## Example

...

## Updates issue/story

ent-3913
ent-3833

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -39,6 +39,7 @@
     "plugin:jsdoc/recommended"
   ],
   "globals": {
+    "mockHook": "readonly",
     "mountHookComponent": "readonly",
     "mockWindowLocation": "readonly"
   },
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
       "!src/index.js",
       "!src/setupTests.js",
       "!src/components/app.js",
+      "!src/components/**/index.js",
       "!src/common/index.js",
       "!src/redux/index.js",
       "!src/redux/store.js",
--- a/public/locales/en-US.json
+++ b/public/locales/en-US.json
@@ -169,7 +169,6 @@
     "tourDescription": "We'll walk you through each step, and include insight into how Red Hat collects and uses subscription data."
   },
   "curiosity-view": {
-    "redirectError": "Redirect failed",
     "title": "{{appName}}",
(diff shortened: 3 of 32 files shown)


.eslintrc
```diff
@@ -39,6 +39,8 @@
     "plugin:jsdoc/recommended"
   ],
   "globals": {
+    "mockReactRouterDom": "readonly",
```
cdcabrera, line 42: required: mockReactRouterDom isn't used, remove

cdcabrera (commented):
Minor update

PR #718, opened 2021-06-23
refactor(chart): ent-3843 context, hooks for chart components

cdcabrera:
## What's included

- fix(useRouter): testing, mockHook to shallowHook
- feat(useWindow): ent-3843 useResizeObserver
   * useWindow, add useResizeObserver, resize event
- refactor(chart): ent-3843 context, hooks for chart components
   * build, use updated Victory components, spelling dictionary
   * chart, chartArea rename/remove, refactor to use context, hooks
   * chartContext, context for related components, toggle handler
   * chartElements, bundle base Victory line, areaChart with props
   * chartHelpers, generate element props
   * chartIcon, return a pf icon, or custom svg
   * chartLegend, container component, provide chart toggle interaction
   * chartTooltip, function, return Victory compatible component, styling
   * graphCard, ChartLegend, Tooltips, update towards new chart components
   * productViews, update towards new chart components

### Notes
- refactors chartArea into multiple chart facets and components, new behavior noted as @mirekdlugosz @ntkathole ...
   - a "tail" should now appear on the tooltip
   - a "dash" icon is now used in the tooltip to represent threshold and match the legend
   - the "infinity" icon is used to represent a threshold of "unlimited" it was determined by UX/design to help optimize space, @bclarhk @rblackbu 
   - legend and tooltip icons should appear slightly larger, we now attempt to use font-awesome icons
   -  when the graph size itself is (NOT screen size) greater than 500px the tooltip is locked at a 75% height along the y axis, and either moves left or right of the vertical cursor line
   - when the graph size itself (NOT screen size) is less than 500px the tooltip allows more movement and attempts to move out of the way of the cursor
   - the order of the tooltip legend items tries to align to the STACKED layer order. there is an exception made for the threshold which does NOT stack
   - we activated some newer Victory aspects around the stacking order for layers. Victory will now attempt to fill in the blanks when it comes to stacked "area charts" 
   - the tooltip itself now has a solid background color instead of the 90%+ opacity from before, this should improve contrast
   - CSS class names were updated
      - @mirekdlugosz @ntkathole if test attributes need to be added to the tooltip container or legend we can add those in, may be a subsequent PR. There was mention of dumping the graph data to a location tests would be able to access, we can review this too as a subsequent update.
- **known issues**
   - ISSUE, the tooltip appears to jump when moving between dates/x coordinations
      - the content of the tooltip is changing size, hence the tooltip attempts to reposition itself. this happens when data sets vary wildly from point to point
   - ISSUE, the vertical cursor line appears over the tooltip
      - Victory appears to have caused an issue with the SVG container portal originally used to help sequence the SVG elements (SVG is reliant of dom stacking order). It becomes a poor user experience when we activate the new "voronoi container"... the tooltip jumps unexpectedly and displays lag. We can approach this again if/when the Victory package updates its container components.

1. `$ yarn test`
1. confirm all tests pass

### Local run check
1. `$ yarn start`
1. confirm the chart is still interactive

### Proxy run check
1. `$ yarn start:proxy`
1. confirm the chart is still interactive

### Check the build
1. `$ yarn build`
1. confirm the build still compiles, and tests pass

## Example

Screen Shot 2021-06-29 at 9 45 54 PM (2)

Screen Shot 2021-06-29 at 11 10 44 PM (2)

## Updates issue/story

ent-3843
attempts to correct pieces of ENT-3690


src/components/graphCard/graphCardChartLegend.js
```diff
@@ -116,8 +90,6 @@ class GraphCardChartLegend extends React.Component {
           position={TooltipPosition.top}
           enableFlip={false}
           distance={5}
-          entryDelay={100}
```
cdcabrera, line 119: optional: may want to consider adding this delay back in, tooltip fires almost immediately in some cases making it difficult to read the graph itself

src/components/chart/chartHelpers.js
```diff
@@ -0,0 +1,403 @@
...
+ * @param {object} params.individualMaxY
+ * @param {number} params.maxX
+ * @param {number} params.maxY
+ * @param {boolean} params.xAxisFixLabelOverlap
+ * @param {number} params.xAxisLabelIncrement
+ * @param {Function} params.xAxisTickFormat
+ * @param {Function} params.yAxisTickFormat
+ * @returns {{xAxisProps: object, yAxisProps: Array}}
+ */
+const generateAxisProps = ({
```
cdcabrera, line 328: optional: plural of Axis is Axes... this wording also works in the singular spelling

cdcabrera (commented):
Need to add in the product view for OpenShift Container, it contains threshold graph config settings

PR #740, opened 2021-07-16
fix(graphCard): ent-3813 header styling for total display

cdcabrera:
## What's included

- fix(graphCard): ent-3813 header styling for total display
- fix(select): focused unit testing

### Notes
- ~**THERE IS AN UNINTENDED SIDE-EFFECT THAT HAPPENS WITH THIS UPDATE**. The more we attempt to fix this the worse it gets, see the below animation. PF select we use is right align, apparently it has issues in the Chrome browser due to this fix. We have to debug this accordingly, and we need to continue the push forward. The last commit on CI for ENT-3813 will be removed temporarily while this work happens.~ @mirekdlugosz 
- ~once approved/verified this will be squashed with the prior related commit~ @mirekdlugosz
- There is an issue around using the "popper" package within PF, we've added updates to compensate
- We've gone ahead and squashed the `overflow styling for header total display` commit from #738 into this PR
- at **around** 580px browser width the CSS **attempts** to left align the contents of "total core hours used" and the select/dropdown lists

### Proxy run check
1. `$ yarn start:proxy`
1. confirm that the OpenShift Console and OpenShift Dedicated "total core hours used" display behaves by attempting to have the copy on a single line, and only falling back to multi-line when the space is limited

### Corrected issue with PF Select

https://user-images.githubusercontent.com/3761375/126174263-22611218-5ce1-4990-afa9-7a530d82ee7d.mp4

#### 580'ish pixels
Screen Shot 2021-07-16 at 3 11 19 PM

#### Multiple sizes and behavior
Screen Shot 2021-07-16 at 3 05 41 PM
Screen Shot 2021-07-16 at 3 05 29 PM
Screen Shot 2021-07-16 at 3 05 11 PM

## Updates issue/story

ent-3813
#738

### DIFF
--- a/src/styles/_usage-graph.scss
+++ b/src/styles/_usage-graph.scss
@@ -9,24 +9,32 @@
       }
 
       &__title {
+        flex-grow: 1;
         padding: 0;
+        padding-right: var(--pf-c-card__actions--PaddingLeft);
       }
 
       &__actions {
+        display: flex;
         margin: inherit;
+        margin-left: auto;
+        padding-left: 0;
+
+        @media (max-width: $pf-global--breakpoint--sm) {
+          margin-left: inherit;
+        }
+
+        & > .curiosity-select {
+          width: auto;
+        }
       }
     }
 
     &__total {
-      min-width: 175px;
-      padding-left: var(--pf-global--spacer--sm);
       padding-right: var(--pf-global--spacer--sm);
       text-align: right;
       white-space: normal;
-
-      @media (min-width: $pf-global--breakpoint--sm) {
-        white-space: nowrap;
-      }
+      word-break: break-word;
     }
 
     &__legend-item {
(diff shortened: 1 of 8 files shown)


src/styles/_usage-graph.scss
```diff
@@ -9,16 +9,28 @@
...
         margin: inherit;
+        margin-left: auto;
+
+        @media (max-width: $pf-global--breakpoint--sm) {
+          margin-left: inherit;
+          padding-left: 0;
+        }
+
+        & > .curiosity-select {
+          display: inline-block;
```
cdcabrera, line 27: optional: looks like pf declares this as "inline-block", shouldn't need this line

cdcabrera:
Appears this update causes a related issue in Chrome around the dropdown/select right alignment, something to do with the width being `auto`, investigating

PR #795, opened 2021-09-16
feat(inventorySubscriptions) ent-3352 activate subscription table

joeldmyers:
## What's included
- Removed dotenv check for product views (RHEL / Openshift) for the subscription inventory table
- confirmed API endpoint is valid
- modified the inventory columns per the storyboard

### Notes
- Table is not currently sortable by title or by the cores/sockets; I tried to update but was getting a 404 back from the API, and it was noted in the story in Jira that "not all inventory column features may be available initially and that it's known to UX/D"
- Also the UOM filtering functionality on the API side is still pending, but with these changes, the UI sorting in Openshift should function correctly.
- Lastly, there is a scenario when filtering by Service Level on RHEL, descending, where UOM is returned as null from the API, resulting in an error in the header; hence why the header config for that has a default value of 'sockets'. 
Screen Shot 2021-09-16 at 10 48 16 AM

## How to test
Proxy run check
update the NPM packages with `$ yarn`
make sure Docker is running, plus on network, then
`$ yarn start:proxy`
navigate to the RHEL product view.
click "Current subscriptions" tab
confirm that data is showing properly
confirm that data is sortable by clicking Service Level, Quantity, and Next Renewal
navigate to the OpenShift product view
click "Current subscriptions" tab
confirm that data is showing properly
confirm that data is sortable by clicking Service Level, Quantity, and Next Renewal
confirm that cores / sockets header and data (once the API has been updated for this) are updated appropriately when clicking the Cores/Sockets select above

## Example
Screen Shot 2021-09-16 at 10 44 32 AM

Screen Shot 2021-09-16 at 10 53 31 AM

...

## Updates issue/story
[ENT-3352](https://issues.redhat.com/browse/ENT-3352)


cdcabrera (changes requested):
Good first round!  Restoring the ability to disable the component should wipe a bunch of updates out too, we'll call it a separation of concerns.

Looks like i18next is automagically finding the closest key without needing to lowercase the response value when you only have a "single" context entry... so looks like we need to still lowercase the value.


**2 additional updates to what you have here should have us up and running**

First, lets add a `case` statement to help lowercase the values

https://github.com/RedHatInsights/curiosity-frontend/blob/ci/src/redux/selectors/subscriptionsListSelectors.js#L111-L114
```
case rhsmApiTypes.RHSM_API_RESPONSE_INVENTORY_SUBSCRIPTIONS_DATA_TYPES.UOM:
  return value?.toLowerCase() || null;
```


Second, lets charge the query for RHEL, we're thinking this is part of the `null` value aspect we're seeing... add in a query for SOCKETS around line `38`, similar to the [`openshiftContainer.js`](https://github.com/RedHatInsights/curiosity-frontend/blob/ci/src/config/product.openshiftContainer.js#L36)

https://github.com/RedHatInsights/curiosity-frontend/blob/ci/src/config/product.rhel.js#L38

```
query: {
  [RHSM_API_QUERY_TYPES.UOM]: RHSM_API_QUERY_UOM_TYPES.SOCKETS,
```

cdcabrera (approved):
lgtm! just rebase and we'll merge it in!

PR #813, opened 2021-10-11
refactor(router): ent-4208 activate routeDetail context

cdcabrera:
## What's included

- refactor(router): ent-4208 activate routeDetail context
   * config, product configs add product label
   * config, routes productId for rhel archs
   * router, hook, add routeDetail context provider
   * hooks, passthrough for useRouteDetail
   * productViews, apply useRouteDetail hook

### Notes
- State management for product configuration, base work for RHOSAK, ent-4366

### Proxy run check
1. `$ yarn start:proxy`
1. navigate between products, confirm existing behavior remains unchanged. there should be no visual or behavioral differences

## Example

...

## Updates issue/story

ent-4208
relates ent-4366


src/components/productView/productView.js
```diff
@@ -185,45 +175,10 @@ const ProductView = ({ routeDetail, t, toolbarGraph, toolbarGraphDescription, to
 /**
  * Prop types.
  *
- * @type {{t: translate, toolbarGraph: (Node|boolean), toolbarGraphDescription: boolean, routeDetail: object,
- *    productConfig: object, toolbarProduct: (Node|boolean)}}
+ * @type {{t: translate, toolbarGraph: (Node|boolean), toolbarGraphDescription: boolean, productConfig: object,
+ *    toolbarProduct: (Node|boolean)}}
  */
 ProductView.propTypes = {
```
cdcabrera, line 181: Optional: since we're removing the propTypes description would be good to confirm the config tests check for these properties

cdcabrera: added productLabel check to existing config pass/fail

PR #826, opened 2021-11-09
fix(routes): ent-4366 update product path id

cdcabrera:
## What's included

- fix(routes): ent-4366 update product path id

### Notes
- switches the product path parameter to `subscription/streams` @mirekdlugosz @ntkathole 
- ~possible alternative to "streams" = "apache-kafka"~ trademark issues
- @zanewoodfin 

### Proxy run check
1. `$ yarn start:proxy`
1. confirm the rhosak display is accessible via `/subscriptions/streams`

## Example

...

## Updates issue/story

ENT-4366


cdcabrera:
Looks like we came across the first issue with using other products to describe products... the missing product view we use groups

- OpenShift Dedicated
- OpenShift Metrics
- OpenShift Container
- and of course OpenShift Streams

This is the warning sign, it may be manageable now... and we can fix the logic for RHOSAK, but the more products we add the more conflicts we'll have. We're tempted to implement some level of product path hash to go around this entirely

PR #832, opened 2021-11-12
fix(helpers): ent-4366 optional chaining for numbro

cdcabrera:
## What's included

- fix(helpers): ent-4366 optional chaining for numbro

### Notes
- convenience for behind the scenes. there are scenarios where the hook based components will "do all the things" to setup and can cause issues with functions expecting strict types
- we also expanded the RHOSAK decimal conversion check using numbro to confirm that as long as we don't receive a decimal places higher it will trim values. In RHOSAKS case we cap at 5 decimal places for all metrics. Instance hours, in theory, should cap itself at 2. 
   - If it doesn't we'll have to go back and create a custom graph mod so that each metric can set its own graph x/y axis ticks if it's standalone. Right now the graph only allows setting the x/y axis ticks as a whole. It should be a similar mod to the x/y graph axis label.
   

1. `$ yarn test`
1. confirm all tests pass. there should be no visual changes associated with this update.

## Example

...

## Updates issue/story

ent-4366


src/common/__tests__/helpers.test.js
```diff
@@ -1,4 +1,4 @@
-import { helpers } from '../helpers';
+import numberDisplay, { helpers } from '../helpers';
```
cdcabrera, line 1: required: misplaced import, remove `numberDisplay`

PR #840, opened 2021-11-19
RHOSAK copy and clean up

cdcabrera:
## What's included

- fix(config): ent-4367 rhosak config checks
- fix(graphCardMetricTotals): ent-4366 daily, monthly header copy
- fix(graphCardMetricTotals): ent-4366 total timestamp formats

### Notes
- using PR as a clean up commit group

#### Timestamp format
Screen Shot 2021-11-19 at 2 12 54 PM

#### Card title updates
Both the "current" and prior months should now display the same card titles
Screen Shot 2021-11-19 at 1 41 15 PM
Screen Shot 2021-11-19 at 1 41 41 PM


jlprevatt (approved):
Part of the screen caps show the old date/time format, but I checked the order of the commits and I think that's why. So looks good.

PR #847, opened 2021-12-02
NPM updates

cdcabrera:
## What's included

- fix(build): npm updates

### Notes
- Upgrading the platform webpack configs appears to cause "issues". Any update >/past the current version of `@redhat-cloud-services/frontend-components-config` of `v4.3.5` ...
   - defaults to a screenwide webpack error display, not clear if it's the webpack one or something custom. Regardless it is REALLY annoying considering the errors come through in both the terminal, console, and platforms error page, and regardless of whether errors or warnings pop up everything is labeled "error". 
   - and second we keep getting a `Could not find "store" in the context of "Connect(Portal)"`... which we thought had to do with notifications. However, removing the notifications platform NPM has little to no bearing on resolving this issue which has the appearance of being a dependency issue, we're not clear what aspect of "rhcsVendor" is causing the issue... see below screenshot
   -    @Hyperkid123 
   - React Router version skipped! React Router has introduced breaking changes in the way routing is handled. 
      - This affects Curiosity a little bit more because we started using context applied through the router route render attribute. Based on our implementation and the changes it looks like we'd have to create wrappers for the routes to maintain the same behavior, doable. But because of the simplistic routing Curiosity uses, the base Router dom hooks we use (and customize) and plan to use, `useHistory` and `useLocation` we do almost as much adjustment to routing with our helpers and existing hooks as the React Router does by itself... why do we need React Router? We are considering rolling our own simplistic router, but we'll need to investigate the complexity of rolling our own vs rolling a wrapper for the new `<Route />` `element` attribute.

## Example
Updating to `@redhat-cloud-services/frontend-components-config`  greater than `v4.3.5` fires this off... when running our local standalone script with `$yarn start`

Screen Shot 2021-12-02 at 5 04 42 PM

## Updates issue/story

ongoing


Hyperkid123:
> defaults to a screenwide webpack error display, not clear if it's the webpack one or something custom. Regardless it is REALLY annoying considering the errors come through in both the terminal, console, and platforms error page, and regardless of whether errors or warnings pop up everything is labeled "error".

That is a webpack-dev-server default. You can turn it off by adding `client: { overlay:false }` to your config.

> and second we keep getting a Could not find "store" in the context of "Connect(Portal)"... which we thought had to do with notifications.

This one is caused by the recent react-redux version. It was causing issues for all applications so we have decided to turn of the module sharing by default. You should be able to fix it by adding it back as a shared module:

```js
  shared: [{ 'react-redux': { requiredVersion: dependencies['react-redux'] } }], // dependencies comes from package.json
```
I don't know the reason yet why some apps require it and some can't use it.

PR #919, opened 2022-04-11
fix(authenticationContext): ent-4669 pass errorStatus correctly

cdcabrera:
## What's included

- fix(authenticationContext): ent-4669 pass errorStatus correctly

### Notes
- Optional chaining was masking the fail, in turn this incorrectly pulled the error status code passing `null`, this in turn caused the OptIn form button to incorrectly display the activate button at multiple points
- @mirekdlugosz @diegomaranhao we probably still need to review what happens after OptIn is submitted

### Proxy run check
1. `$ yarn start:proxy`
1. Confirm a user needs to be Opted in, a 403 should be generated by any RHSM API endpoint
   - Step through the Optin process, the button should be active
   - Next click the OptIn button to activate
   - A toast notification should appear in the upper right along with disabling the button
   - Hitting refresh may bring up the OptIn form again with a disabled OptIn button that reads "Subscriptions active", OR there is a possibility a productView may display related to API submission

## Example

...

## Updates issue/story

ent-4669

### DIFF
--- a/src/components/authentication/__tests__/authenticationContext.test.js
+++ b/src/components/authentication/__tests__/authenticationContext.test.js
@@ -17,8 +17,10 @@ describe('AuthenticationContext', () => {
             locale: {}
           },
           responses: {
-            errors: {
-              status: 403
+            id: {
+              errors: {
+                status: 403
+              }
             }
           }
         })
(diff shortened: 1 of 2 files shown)


mirekdlugosz (approved):
Yup, I should be able to verify that as soon as it's deployed to stage-beta. Plus we will have full regression results tomorrow.

PR #931, opened 2022-05-12
PR for testing app-sre integration

mirekdlugosz:
This is not a real PR. But we need something to test integration with app-sre infra now that app-interface MR is merged.


mirekdlugosz:
I would keep it as workaround for one problem we still have with app-sre infra / ephemeral environments.

* You need container to deploy to EE, e.g. when you want to test deployment.
* app-sre maintains container images build from `main`/`master` branch - these are never removed. Containers build from PRs are removed automatically after few days.
* We haven't yet merged app-sre scripts into `main`, so we don't have "stable" container.
* Which means that you have to create some PR to build container image if you want to play with EE.

As long as we keep good pace of PRs, it doesn't matter. But if we haven't opened anything in last few days, you need fake PR just for container building. I would keep **this** PR for this exact purpose.

But if you prefer to close it, I'm fine with that.

mirekdlugosz:
Since #952 , files required by app-sre infrastructure are available in main branch. It means that quay now contains non-expiring curiosity containers. They may be used for basic EE-related testing, and this PR is no longer needed.

PR #937, opened 2022-05-25
chore(testing): ephemeral envs, skip dist check

cdcabrera:
## What's included

- chore(testing): ephemeral envs, skip dist check
- chore(testing): allow skipping individual tests
- build(dev,proxy,prod): remove unused html template

### Notes
- Cleans up unused HTML template and webpack config
   - With updates to @redhat-cloud-services/frontend-components-config v4.6.9+ we no longer need to maintain our own HTML template, we updated in PR #930 
- Skips the integration dist check associated with ephemeral envs.  @mirekdlugosz 
   - We're having to make an educated guess that the containerized build is applying a base image which in turn causes the build output to alter. It's not great disabling a confirmation check we've used for the past few years but that check is still going be allowed to run against our GitHub actions... just removing some duplication in effort...
   - That segues into the level of repetition in testing that's happening between GitHub actions, Travis, and containerized checks. We'll need to review how much we really need to duplicate in our ephemeral test checks if GitHub actions is also running. This is something we started to do for the Travis facet but decided to just let it provide duplicated effort.
   - GitHub actions is still our preferred path for running consistent, fast, multi-nodejs dev level unit tests and integration checks.

### Confirm bot checks
1. Confirm the `ci.ext.devshit.net PR build` bot is passing

## Example

...

## Updates issue/story

Ongoing


cdcabrera (commented):
minor update to package.json

PR #951, opened 2022-06-24
fix(rhosak): ent-4689 (sw-93) billing provider, remove none option

cdcabrera:
## What's included

- ~fix(rhosak): ent-4689 remove none option~ fix(rhosak): sw-93 billing provider, remove none option

### Notes
- removes the `none` option, behavior for #936  should be updated to reflect this
- introduces "unique" additional options for our dropdown listing components when compared with existing lists
   - as silly as this seems, this update will affect generating future lists, and force unique qualities to be configured either in the API or GUI
      - consistency of dropdown options vs what the API provides is the underlying issue... why is the option being provided in the first place
- ~marking this as blocked since "removing none" was posed as a solution without accounting for configuration~

### Proxy run check
1. `$ yarn start:proxy`
1. navigate to RHOSAK, confirm the select/dropdown only has the options displayed in the below example

## Example

NONE-REMOVED-Screen Shot 2022-06-24 at 12 22 27 AM

## Updates issue/story

ent-4689 aka SWATCH-93

### DIFF
--- a/src/components/toolbar/__tests__/__snapshots__/toolbarFieldBillingProvider.test.js.snap
+++ b/src/components/toolbar/__tests__/__snapshots__/toolbarFieldBillingProvider.test.js.snap
@@ -12,11 +12,6 @@ Array [
     "title": "t(curiosity-toolbar.billing_provider, {\\"context\\":\\"aws\\"})",
     "value": "aws",
   },
-  Object {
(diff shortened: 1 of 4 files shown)


bclarhk:
Yep, remove none plz thx

PR #958, opened 2022-07-27
fix(build): sw-460 npm updates

cdcabrera:
## What's included

- fix(build): sw-460 npm updates

1. `$ yarn test`

### Local run check
1. `$ yarn start`
1. confirm this fails, this is being handled via PR #957 

### Proxy run check
1. `$ yarn start:proxy`
1. confirm you can log in and view the application and subsequent products

### Check the build
1. `$ yarn build`
1. confirm the build completes successfully and without error

## Example

...

## Updates issue/story

sw-460
related to #957
related to #954


src/components/pagination/__tests__/__snapshots__/pagination.test.js.snap
```diff
@@ -65,7 +66,7 @@ exports[`Pagination Component should render a non-connected component: non-conne
       "items": "",
       "itemsPerPage": "Items per page",
       "ofWord": "of",
-      "optionsToggle": "Items per page",
+      "optionsToggle": "",
```
cdcabrera, line 69: optional/future:
we may have to set this prop manually now, confusing why PF decided to not continue pulling placeholder like content for this. keep an eye on it

cdcabrera (commented):
optional/future:
looks like local run with `start:proxy` highlights an icon misalignment on the graph legend... basically the pretty squares don't align veritcally. We've seen this before with PF updates, and it does appear consistent with what is currently in `prod`. Sometimes it gets corrected by moving stuff into environment, missing platform stylesheet locally, etc.

keep an eye on it... if it bugs the designers we'll need to try aligning them
Screen Shot 2022-07-27 at 11 56 14 AM

@bclarhk

cdcabrera (commented):
There's also a future local run set of warnings via Victory charts where they attempted to include source maps for local dev, but still managed to include the refs in their NPM dist... we're ignoring the warnings and continuing the update, but heads up...

related
- https://github.com/patternfly/patternfly-react/issues/7614
- https://github.com/FormidableLabs/victory/pull/2307

cdcabrera:
As of 20220727 we need to actively setup a PR for moving to NodeJS v16 there are multiple resources who now have a minimum of NodeJS v14.x.x

PR #989, opened 2022-10-17
3 of 4 - refactor(router): sw-625 move to router v6

cdcabrera:
## What's included

- refactor(router): sw-625 move to router v6

### Notes
- the routing move and related PRs 
   - create a generated list of pre-sorted product configs on app load. this is prep for loading actual product configurations
   - eliminate the need to have a large independent routing configuration that defines every route. instead we now use a route parameter see `:productPath`. This parameter is "kind-of" superfluous due to the use of a noticeably quicker way of loading product configuration by handling `window.location.pathname` directly within the new `routerContext.js` hooks and `routeHelpers.js`
   - removes the use of react context in favor of simple hooks that
      - initialize the product config lookup with `useSetRouteDetail` and
      - provide product configuration based on a route parameter/config lookup, using product configuration details and/or their closest match based on pathname... 
      - partial product names typed into the url will return a product now. The caveat is, it's what the GUI/algorithm determines to be the closest string match which doesn't always align to "human". This can be disabled if we determine it's causing problems
   - simplifies components, code... always good 

1. `$ yarn test`

### Check the build
1. `$ yarn build`

## Example

...

## Updates issue/story

swatch-625

related #1048 #1045


cdcabrera:
As of 20221207 rebased and updated. Confirmed working both locally, and through the proxy against stage-beta.

We'll need to optimize how the route configs are being pulled since there looked like lag coming through the page loads, but that could also be associated with the proxy. The route configs don't once change so caching them on an extended cache cycle might become a thing

src/components/productView/productView.js
```diff
@@ -84,14 +103,82 @@ const ProductView = ({ t, useRouteDetail: useAliasRouteDetail }) => {
       </ProductViewContext.Provider>
     );
   };
+  */
+  const renderProduct = useCallback(() => {
```
cdcabrera, line 107: required:
confirm this is necessary when compared against the useRouteDetail hook

src/components/i18n/i18n.js
```diff
@@ -24,35 +24,38 @@ const I18n = ({ children, fallbackLng, loadPath, locale }) => {
...
-          defaultNS: 'default',
-          react: {
-            useSuspense: false
-          }
-        });
-    } catch (e) {
-      //
-    }
+    console.log('>>> mount i18n');
+    if (!initialized) {
```
cdcabrera, line 28: required:
confirm this is doing something

cdcabrera: partially confirmed, we'll remove it. appears as a leftover from env testing

src/components/productView/productView.js
cdcabrera: inconclusive, we're going to leave it in for now and reevaluate. it's not hurting us currently. the odd random double/triple load makes it difficult to determine when using the proxy

PR #1008, opened 2022-11-23

mirekdlugosz:

### DIFF
--- a/deploy/pr_check.sh
+++ b/deploy/pr_check.sh
@@ -15,22 +15,55 @@ COMMON_BUILDER=https://raw.githubusercontent.com/RedHatInsights/insights-fronten
 # --------------------------------------------
 # Options that must be configured by app owner
 # --------------------------------------------
-IQE_PLUGINS="curiosity"
-IQE_MARKER_EXPRESSION="smoke"
-IQE_FILTER_EXPRESSION=""
+export IQE_PLUGINS="curiosity"
+export IQE_IMAGE_TAG="curiosity"
+export IQE_MARKER_EXPRESSION="ephemeral"
+export IQE_FILTER_EXPRESSION=""
+export IQE_ENV="ephemeral"
+export IQE_SELENIUM="true"
+export IQE_CJI_TIMEOUT="30m"
 
+# --------------------------------------------
+# Run unit tests, build container and push it to quay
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
OK, so this passed. If you go to "details" under ci.int.devshift.net PR build and log in, you can click "Test Results" and see breakdown of tests that were run. Which is nice.

I am pretty sure accessing ci.int.devshift.net requires VPN.

Some time breakout of this specific run:

- running unit tests, building container, pushing it to quay: 02:49
- deploying new ephemeral environment with subscriptions, including backend and dependencies: 10:31
- deploying pod with QE tests and running them: 02:06
- the whole thing: 16:08

GitHub tells me GitHub action took 8 minutes, so App SRE workflow doubles that.

For the reference, backend team needs around 20 minutes to complete their PR job on App SRE infra. We'll probably need a little more, because UI tests are always going to be slower than API. Obviously I don't want to make that too slow, I think 25 minutes is reasonable, 30 minutes is absolute maximum. We will have to cut the scope of UI tests to meet this time requirement - we will run fewer tests than backend team does.

cdcabrera:
@mirekdlugosz would removing the redundancy of running the GUI tests make much of a difference (removing the [`$ yarn test:ephemeral` and related node scripts in `package.json`](https://github.com/RedHatInsights/curiosity-frontend/blob/main/package.json#L58))? 

If we haven't already removed the redundant unit test checks, It would be preferable if we could and focus on just QE's facet.

Most of it is we still use GitHub actions for
- not having to login to review unit test results on a public project
- codecov aspect of actions, unclear if this support works or is even provided with the behind-the-scenes checks. we generate the coverage, just unclear where it goes?
- a more in-depth aspect of dev integration checks. we had to disable parts of those checks to get them to pass for the behind-the-scenes checks
- commit linting
- and generally just easier to find documentation =)

cdcabrera (approved):
Posted a non-blocking question/comment. Just need a rebase and we should be good

cdcabrera:
As of 20221130 holding PR. It's recommend we hold until there are more tests associated with this check

san7ket:
Closing this in favor of https://github.com/RedHatInsights/curiosity-frontend/pull/1072

PR #1020, opened 2022-12-08
feat(graphCardChart): sw-690 config, action toolbar

cdcabrera:
## What's included

- feat(graphCardChart): sw-690 config, action toolbar

### Notes
- restores the granularity filters for RHEL, OpenShift, and Satellite to their original positions above the graph card
- creates a configurable toolbar for graph card actions
   - allows a configurable area for standalone metric graph cards

## Example

Screen Shot 2022-12-08 at 1 51 43 AM
Screen Shot 2022-12-08 at 1 51 29 AM

#### Configurable metric card toolbar
- _currently form fields placed in the card context are not limited by the metric context, this would require future work_
Screen Shot 2022-12-12 at 2 07 24 PM

## Updates issue/story

sw-690
relates sw-235


src/components/graphCard/graphCardContext.js
```diff
@@ -145,10 +147,58 @@ const useGetMetrics = ({
...
+  useMetricsSelector: useAliasMetricsSelector = useMetricsSelector,
+  useGraphCardContext: useAliasGraphCardContext = useGraphCardContext
+} = {}) => {
+  const [updatedActions, setUpdatedActions] = useState();
+  const { pending, dataSets } = useAliasMetricsSelector();
+  const { settings = {} } = useAliasGraphCardContext();
+  const { actionDisplay, actions = [] } = settings;
+
+  useShallowCompareEffect(() => {
+    if (!pending && dataSets) {
```
cdcabrera, line 170: required: review the pending and dataSets condition

something similar was removed from the metrics call in #1023

cdcabrera: * converted to useMemo since we're dealing with settings
* pending and dataSets length used to limit "content" callbacks. 
   * in the future we may expand this to include a `typeof content === function` check. The limitation associated with using dataSets is the callback

PR #1027, opened 2022-12-15
refactor(select): sw-746 select wrapper class to func

cdcabrera:
## What's included

- refactor(select): sw-746 select wrapper class to func

### Notes
- there are no visual changes associated with this update. HOWEVER, this does affect all product views with select/dropdown fields
   - @mirekdlugosz it would be good to specifically check for the `data-test` attribute. We had to modify how we append it based on PF updates

- part of a base set of refactors to help with upcoming design updates
- refactor/perf update for the pf select wrapper. the current select wrapper has some older ways of doing things, and it can be slightly sped up with some of its cycling

1. `$ yarn test`

### Proxy run check
1. `$ yarn start:proxy`
1. this update affects ALL product views and ALL selects across Subs
   - confirm that select behavior is working as intended
   - data-test attribute used by QE needs confirmation as not having changed. Unfortunately we had to update some of how this is implemented

### Check the build
1. `$ yarn build`
1. confirm build checks come back clean

## Example

...

## Updates issue/story

sw-746


cdcabrera:
As of 20221215 this is a placeholder for updates to the design work, starting with a refactor of the select and table components with updated PF4 components, such as the composable table

src/components/form/select.js
```diff
@@ -1,80 +1,364 @@
...
+ * @param {*|React.ReactNode} params.selectField
+ * @param {object|Array} params.options
+ * @param {string|number|Array} params.selectedOptions
+ * @param {string} params.variant
+ * @param {object} params.props
+ * @returns {{options: *[]|*, selected: *[]}}
  */
-class Select extends React.Component {
-  state = { isExpanded: false, options: null, selected: null };
+const formatOptions = ({ selectField = { current: null }, options, selectedOptions, variant, ...props } = {}) => {
```
cdcabrera, line 80: optional: review using a memoize on this format helper... it is understood that updating "selectedOptions" might make it useless

cdcabrera: looks like we already did a review on memoize, the notes indicate the gain was minimal, a human wouldn't notice the difference... but we'll check again

cdcabrera: the state update combined with the effect diff help prevent unnessary use... adding in the memoize, again, provides no noticeable gain other than a possible head scratch

src/components/form/__tests__/__snapshots__/select.test.js.snap
```diff
@@ -631,19 +664,20 @@ exports[`Select Component should render an expanded select: expanded 1`] = `
 ]
 `;
 
-exports[`Select Component should return an emulated onchange event: checkbox emulated event 1`] = `
+exports[`Select Component should return an emulated onchange event, checklist variant: checklist emulated event 1`] = `
```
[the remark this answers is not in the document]
cdcabrera: appears to be returning correctly

src/components/form/select.js
```diff
@@ -1,80 +1,365 @@
...
+  const { current: domElement = {} } = selectField;
+  const dataAttributes = Object.entries(props).filter(([key]) => /^data-/i.test(key)); // eslint-disable-line
```
cdcabrera, line 83: required: this line is unnecessary, clean up from testing

PR #1066, opened 2023-02-22
build(dev,proxy,prod): sw-917 config beta to preview

cdcabrera:
## What's included

...

### Notes
- "beta" is used everywhere
   - need to confirm branch naming on the current build repo will be maintained since we're moving to the container workflow
      - need to confirm this affects staging as well... it may be awkward to change production only
   - we use a dotenv setting that helps set the `beta` prefix
   - we're leaving the `BETA` copy in repo, this affects the variable setting
   - the build will need to be confirmed as working in
      - lower envs
      - production envs
- this may affect sw-625 our version 6 routing updates, and cause a potential subsequent patch 
- this will affect sw-817 our documentation effort
   - the CONTRIBUTING.md updates need to be "more generic" when it describes alternate envs

## Example

...

## Updates issue/story

sw-917

### DIFF
--- a/config/webpack.proxy.config.js
+++ b/config/webpack.proxy.config.js
@@ -10,7 +10,7 @@ const { _BUILD_RELATIVE_DIRNAME, DEV_BRANCH, DEV_PORT } = setupDotenvFilesForEnv
 let BETA_PREFIX = '';
 
 if (/(prod|stage|qa|ci)-beta/.test(DEV_BRANCH)) {
-  BETA_PREFIX = '/beta';
+  BETA_PREFIX = '/preview';
 }
 
 const { config: webpackConfig, plugins } = config({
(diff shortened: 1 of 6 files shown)


config/webpack.proxy.config.js
```diff
@@ -10,7 +10,7 @@ const { _BUILD_RELATIVE_DIRNAME, DEV_BRANCH, DEV_PORT } = setupDotenvFilesForEnv
 let BETA_PREFIX = '';
 
 if (/(prod|stage|qa|ci)-beta/.test(DEV_BRANCH)) {
```
cdcabrera, line 12: missed this `beta` reference

cdcabrera: this is fine, we're leaving references to the envs as `beta`... this won't be confusing at all

cdcabrera:
As of 20230426 closing this down, it's too much based on the current way consoledot is handling the redirect. Altering the entirety of `beta` breaks local development, instead a lighter version is here #1109

PR #1071, opened 2023-03-01
fix(productViewMissing): sw-625 local dev view, run

cdcabrera:
## What's included

- fix(productViewMissing): sw-625 local dev view, run

### Notes
- non blocking local developer workflow view and run
- we ignored/disabled the local developer workflow to get the router updates out. this pr reactivates it 
- this may, or may not, throw an error in console in environment since routing is now tiered/nested with consoledot

### Local run check
1. `$ yarn start`
1. confirm that the default card display appears and can be used to navigate between all available products, see example
   - _note: the selected card attempts a url path closest match for product_

### Proxy run check
1. `$ yarn start:proxy`
1. confirm routing/navigating between products remains unchanged
   - confirm manually navigating towards a url/path under subscriptions like `[consoledot path]/subscriptions/lorem-ipusm/rhacs` redirects to the closest product match without throwing a console error.
   - _note: an end-user should never be able to directly receive the developer multi-card-product view_

## Example

Screen Shot 2023-03-02 at 4 18 16 PM

### DIFF
--- a/src/components/productView/productViewMissing.js
+++ b/src/components/productView/productViewMissing.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useLayoutEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Button, Card, CardBody, CardFooter, CardTitle, Gallery, Title, PageSection } from '@patternfly/react-core';
 import { ArrowRightIcon } from '@patternfly/react-icons';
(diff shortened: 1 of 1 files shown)


cdcabrera:
20230228 - 20230301 blocking due to env issues. Unable to confirm this fix is working

PR #1077, opened 2023-03-09
refactor(config): sw-922 product custom card metrics

cdcabrera:
## What's included

- refactor(config): sw-922 product custom card metrics

### Notes
- this **affects the following on-demand displays**; RHOSAK/streams, RHODS, RHACS
- moves the hardcoded aspects of metric cards onto product configuration
   - allows the use of shared metric cards under general graph settings
   - allows the use of override metric cards at the filter settings level
   - allows independent metric card disabling  
- leverages the newer ability to place `testId`s against locale strings... RHOSAK/streams, RHODS, RHACS now use the following identifiers against data
   - `graphDailyTotalCard-header` (description area)
   - `graphDailyTotalCard-body` (numeric/total area)
   - `graphDailyTotalCard-footer` (date area)
   - `graphMonthlyTotalCard-header` (description area)
   - `graphMonthlyTotalCard-body` (numeric/total area)
   - `graphMonthlyTotalCard-footer` (date area)

1. `$ yarn test`

### Proxy run check
1. `$ yarn start:proxy`
1. confirm that the following products have their metric card displays "unchanged"
   - RHACS
   - RHODS
   - RHOSAK/streams

## Example

...

## Updates issue/story

sw-922 sw-923

### DIFF
--- a/src/components/graphCard/__tests__/__snapshots__/graphCardMetricTotals.test.js.snap
+++ b/src/components/graphCard/__tests__/__snapshots__/graphCardMetricTotals.test.js.snap
@@ -1,9 +1,111 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`GraphCardMetricTotals Component should handle custom card displays: fulfilled 1`] = `
+<div
+  className="curiosity-usage-graph__totals"
+  data-test="graphMetricTotals-hello-world"
+  data-test-data="{"groupMetricId":["hello","world"]}"
+>
+  <div>
+    <div
+      className="curiosity-usage-graph__totals-column"
+    >
+      <Card
+        className="curiosity-usage-graph__totals-column-card "
+        data-test="graphMetricTotalsCard-0"
+        isPlain={true}
+        key="3409dd31007d6ac3d52cd7bf810296df8201c75d"
+      >
+        <CardHeader>
+          <CardTitle>
+            <Title
+              headingLevel="h2"
+              size="md"
+            >
+              lorem
+            </Title>
+          </CardTitle>
+        </CardHeader>
+        <CardBody>
+          <div>
+            t(ipsum-{{data}}, {"testId":"custom test id","data":"{\\"groupMetricId\\":[\\"hello\\",\\"world\\"]}"})
+          </div>
+        </CardBody>
+        <CardFooter>
+          <div>
+            dolor sit timestamp
+          </div>
+        </CardFooter>
+      </Card>
+      <Card
+        className="curiosity-usage-graph__totals-column-card "
+        data-test="graphMetricTotalsCard-1"
+        isPlain={true}
+        key="my custom react key"
+      >
+        <CardHeader>
+          <CardTitle>
+            <Title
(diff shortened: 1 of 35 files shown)


src/components/graphCard/graphCardMetricTotals.js
```diff
@@ -54,134 +52,110 @@ const GraphCardMetricTotals = ({
   const dailyHasData = isCurrent ? currentHasData : lastHasData;
   const dailyValue = isCurrent ? currentValue : lastValue;
 
-  return (
-    <div data-test={`graphMetricTotals-${_camelCase(metricId)}`} className="curiosity-usage-graph__totals">
```
cdcabrera, line 58: required: the `data-test` attributes are used by QE we need to make sure an alternative setup is in place before removing these attributes

mirekdlugosz: That is correct, iqe-curiosity-plugin depends on `graphDailyTotalCard` attribute value.

I see below that there's still `data-test`, but value is `graphCard-${index}`. Assuming index evaluates to numerical index, I don't really fancy this idea. I would prefer to have descriptive names - `graphCard-daily`, `graphCard-overage` etc.

Changing string itself is alright, but will require us to change string in iqe plugin, too. We might have a regression run or two failed between the time this is deployed and our automation is fixed, depending on how well we coordinate this.

cdcabrera: review updates are in
- we're retaining the parent `data-test` attribute... it's located above the cards, line `58` in this set of comments, `graphMetricTotals-${_camelCase(metricId)}``
   - this metricId will get "fuzzy" in the future when more than one metric is displayed in the graph that is dissimilar... ie. Cores vs Sockets. This could be resolved with something like `graphMetricTotals-${_camelCase(metricId-1)}-${_camelCase(metricId-2)}-${_camelCase(metricId-3)}`
- unfortunately, the child cards are still going to be hit if we want to make the cards configurable based on anything ux/design wants in there. Which means the current solution `graphMetricTotals-${index}` is where we're stuck. However, there is one possible alternative we can think of, we're sure there are more out there...
   - we pass in a specific `setting` in product configuration solely for testing purposes... not a big fan of a singular purpose setting, but it is feasible @mirekdlugosz  
   - we are def open to alternative solutions

cdcabrera: forgot to also note
- the current limit on nested metric cards being planned is `3` to `4`... so `index` would never go past that set point

src/components/graphCard/__tests__/__snapshots__/graphCardMetricTotals.test.js.snap
```diff
@@ -247,61 +154,4 @@ exports[`GraphCardMetricTotals Component should handle multiple display states:
 </div>
 `;
 
-exports[`GraphCardMetricTotals Component should render a basic component: basic 1`] = `
```
cdcabrera, line 250: required: restore some level of the `basic` snapshot

PR #1112, opened 2023-04-27
build(prod,stage): sw-917 dotenv config for surface urls

cdcabrera:
## What's included

- build(prod,stage): sw-917 dotenv config for surface urls

### Notes
- The update on prod from `beta` to `preview` on 20230426 highlighted a missing "hardcoded-like" path for multiple urls. Curiosity "shares" the `beta` prefix through dotenv files making `beta`, and now `preview` available during compile for both the build and GUI
   - Curiosity uses a "path prefix" to load remote locale strings. This update moves and applies "preview" to that path instead of beta while still deploying to "beta/apps"
   - Curiosity uses a link back to inventory displays, this is also affected

### Check the build
1. `$ yarn build`

### Check the preview redirect, and beta env
1. navigate towards preview staging `subscriptions`
   - open the dev console and confirm `en-US.json` reflects the correct path
   - confirm the inventory links point at "preview"
1. navigate towards beta staging `subscriptions`
   - open the dev console and confirm `en-US.json` reflects the correct path
   - confirm the inventory links point at "preview" (we're doing this now to avoid updates later since the intention is to make `preview` the primary path)

## Example

...

## Updates issue/story

sw-917

### DIFF
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -11,7 +11,7 @@ const {
 
 const { config: webpackConfig, plugins } = config({
   rootFolder: _BUILD_RELATIVE_DIRNAME,
-  deployment: (/beta/.test(BETA_PREFIX) && 'beta/apps') || 'apps',
+  deployment: (/beta|preview/.test(BETA_PREFIX) && 'beta/apps') || 'apps',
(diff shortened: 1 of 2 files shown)


cdcabrera:
This update failed to correct the issue, see #1113

PR #1113, opened 2023-04-27
build(prod,stage): sw-917 dotenv config for preview urls

cdcabrera:
## What's included

- build(prod,stage): sw-917 dotenv config for preview urls

### Notes
- this update will be squashed into PR #1112 to avoid confusion in the future
- #1112 failed to work for strings, but does work for urls
   - if we want "preview" to show up in the shortterm for links but not break resources still hosted under `beta` this update is unfortunately necessary
   - we can roll this change back or alias the new new dotenv `UI_DEPLOY_PATH_LINK_PREFIX` param so if someone decides to change from "preview" to "whatever" in the future we're set

### Check the preview redirect, and beta env
1. navigate towards preview staging `subscriptions`
   - open the dev console and confirm `en-US.json` reflects the correct path
   - confirm the inventory links point at "preview"
1. navigate towards beta staging `subscriptions`
   - open the dev console and confirm `en-US.json` reflects the correct path
   - confirm the inventory links point at "preview" (we're doing this now to avoid updates later since the intention is to make `preview` the primary path)

## Example

...

## Updates issue/story

sw-917
fixes #1112


cdcabrera:
Confirmed on stage, this appears to correct both issues
- locale strings
- inventory links

PR #1153, opened 2023-07-12
build(release): frontends jenkinsfile pr checks

maknop:
WIP for testing Jenkinsfile pipeline.


cdcabrera:
@maknop just a heads up. This week of 20230725 we're sliding the Jenkins file commits around as we do releases since `main` is our primary branch. This commits will still be there, but the hashes may update and cause some merge conflicts to pop up

PR #1194, opened 2023-09-19
1 of 3 PF5 migration, refactor(build): sw-1689 pf5 updates

cdcabrera:
## What's included

- refactor(build): sw-1689 pf5 updates

### Notes
- pf5 updates only, the optimized build will need to be completed in a subsequent PR
   - our build makes use of dotenv params which require customization of the consoledot ui cli tool
   - initial check to update the build required more modification than expected
- there are minor spacing/padding differences associated with PF4/5
- the curiosity select component wrapper continues to make use of the deprecated dropdown and select components since it solved the issue of multiple components doing similar things. this will need to be migrated to pf5 in a subsequent PR

1. `$ yarn test`

### Check the build
1. `$ yarn build`

## Example

...

## Updates issue/story

sw-1689


src/styles/_icon.scss
```diff
@@ -0,0 +1,59 @@
...
+          > svg {}
+        }
+      }
+       */
+    }
+  }
+
+  span.curiosity-icon__question {
+    display: inline-block;
+    padding-left: var(--pf-v5-global--spacer--sm);
```
cdcabrera, line 51: required: centralize the `pf-5` css variables so we can avoid this type of update in the future

PR #1264, opened 2024-01-30
fix(productView): key causes refresh

cdcabrera:
## What's included

- fix(productView): key causes refresh
- ~refactor(productView): simplify, unnecessary hook~
   - this commit can be found here, https://github.com/cdcabrera/curiosity-frontend/tree/20240130-productview-idealsolution

### Notes

- ~simplifies the product view component by removing the no longer necessary `useCallback`~
   - ~this was done because of routing overlap between `swatch` and `subscriptions inventory`~
- ~align to authentication component, consistency~
- this started as a `fix` for the `key` being used on the provider... we realized unnecessary view refreshes were happening because the key aligned to `productID` instead of `productGroup`... updating to `productGroup`  removed the refresh issue (and removing also works)... then we realized the hook itself might not be necessary...
   - UPDATE, the original `useCallback` is still necessary after testing... the original routing configs are still being applied in production... the ideal fix is here once consoledot corrects the routing... https://github.com/cdcabrera/curiosity-frontend/tree/20240130-productview-idealsolution

1. `$ npm test`

### Check the build
1. `$ npm run build`

1. `$ npm test`

### Local run check
1. `$ npm start`
1. confirm the productView loads successfully and product variants can be navigated as intended

### Proxy run check (FAILED, reverting to lesser update fix instead, intended change is linked above)
1. `$ npm run start:proxy`
1. confirm the productView loads successfully and product variants can be navigated as intended
1. CONFIRM that navigating between other Subscription applications, specifically Inventory, does not cause undue page refreshes due to routing overlap... this is an issue we faced in the past due to the ORIGINAL ROUTING (now that we're under `usage` the issue may no longer exist... but if someone forgot to remove the old routing configs...)

### Check the build
1. `$ npm run build`

## Example

...

## Updates issue/story

ongoing

### DIFF
--- a/src/components/productView/productView.js
+++ b/src/components/productView/productView.js
@@ -44,7 +44,7 @@ const ProductView = ({ t, useRouteDetail: useAliasRouteDetail }) => {
       }
 
       return (
-        <ProductViewContext.Provider value={config} key={`product_${productId}`}>
+        <ProductViewContext.Provider value={config}>
           <PageMessages>
             <BannerMessages />
           </PageMessages>


cdcabrera:
As of 20240130 holding due to environment testing issues

PR #1278, opened 2024-02-21
feat(serviceConfig): sw-2024 limit, expire poll

cdcabrera:
## What's included

- feat(serviceConfig): sw-2024 limit, expire poll

### Notes
- add an poll limit to help avoid runaway timers

## Example

...


cdcabrera:
As of 20240228 closing this down. 

The concept works, but after discussion it's adding additional complexity for a feature that's technically part of the "validator", the `retry count`, which is included with the polling. The only difference which comes out the same in the end is the user has to set the limit vs the expiration which is calculated as `total time` / `poll interval` = `retry limit`

PR #1287, opened 2024-03-04
test(serviceConfig): sw-2256 axios param checks

cdcabrera:
## What's included

- test(serviceConfig): sw-2256 axios param checks

### Notes
- cover axios implementation of `undefined` and `null` query parameters
   - the gui is reliant on this feature to help in resets

## Example

...

### DIFF
--- a/src/services/common/__tests__/__snapshots__/serviceConfig.test.js.snap
+++ b/src/services/common/__tests__/__snapshots__/serviceConfig.test.js.snap
@@ -616,3 +616,36 @@ exports[`ServiceConfig should have specific properties and methods: specific pro
   "globalResponseCache",
 ]
 `;
+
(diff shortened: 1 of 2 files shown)


cdcabrera:
so this test doesn't quite work like we thought, even after installing the updated axios package it still returns as a match/true, continuing to see if we can pull something out of the service config to check.

cdcabrera:
this test continues to produce a false positive... in order to fully be able to test this we have to mock so much of axios its become beyond scope... something with parameter serialization in `0.28` is busted. 

we have 2 paths
- do the quick triage of ignoring this test, since updating to axios 1+ seems to not present the same problem, and hope it doesn't pop back up in a next version
- or try using axios config custom `paramsSerializer` option so we can maybe intercept the issue in the future

cdcabrera:
closing in favor of #1288 followed by an additional pr that breaks more unit tests for our axios configuration

PR #1302, opened 2024-03-26
fix(inventoryLinks): sw-2326 use inventory_id instead of instance_id

Sgitario:
## What's included
In [SWATCH-2144](https://issues.redhat.com/browse/SWATCH-2144), we changed the existing instance_id from inventory_id to the actual provider id. However, the UI was using this instance_id to build up the link to the inventory service. So, we need to provide the inventory_id, so the UI can continue building this url.

Related to https://github.com/RedHatInsights/rhsm-subscriptions/pull/3188.

## How to test
1.- `npm start`
2.- browse to `http://localhost:3000/apps/subscriptions/openshift`
3.- when moving the cursor over any item name, the link should look like `http://localhost:3000/insights/inventory/CCCCC-b344-4778-831c-CCCCCCC/` instead of `http://localhost:3000/insights/inventory/i-4778-831cCCCCCCC/`

I've also verified that the links for ACM (here, we need to keep using `instance_id`), OpenShift Metrics and OpenShift Dedicated works. 

## Updates issue/story
Jira issue: [SWATCH-2326](https://issues.redhat.com/browse/SWATCH-2326)

### DIFF
--- a/src/config/product.openshiftContainer.js
+++ b/src/config/product.openshiftContainer.js
@@ -194,18 +194,18 @@ const config = {
       cell: (
         {
           [INVENTORY_TYPES.DISPLAY_NAME]: displayName,
-          [INVENTORY_TYPES.INSTANCE_ID]: instanceId,
+          [INVENTORY_TYPES.INVENTORY_ID]: inventoryId,
           [INVENTORY_TYPES.NUMBER_OF_GUESTS]: numberOfGuests
         } = {},
         session
       ) => {
         const { inventory: authorized } = session?.authorized || {};
 
-        if (!instanceId) {
+        if (!inventoryId) {
           return displayName;
         }
 
-        let updatedDisplayName = displayName || instanceId;
+        let updatedDisplayName = displayName || inventoryId;
 
         if (authorized) {
           updatedDisplayName = (
@@ -213,7 +213,7 @@ const config = {
               isInline
               component="a"
               variant="link"
-              href={`${helpers.UI_DEPLOY_PATH_LINK_PREFIX}/insights/inventory/${instanceId}/`}
+              href={`${helpers.UI_DEPLOY_PATH_LINK_PREFIX}/insights/inventory/${inventoryId}/`}
             >
               {updatedDisplayName}
             </Button>
--- a/src/config/product.openshiftMetrics.js
+++ b/src/config/product.openshiftMetrics.js
@@ -224,18 +224,18 @@ const config = {
       cell: (
         {
           [INVENTORY_TYPES.DISPLAY_NAME]: displayName,
-          [INVENTORY_TYPES.INSTANCE_ID]: instanceId,
+          [INVENTORY_TYPES.INVENTORY_ID]: inventoryId,
           [INVENTORY_TYPES.NUMBER_OF_GUESTS]: numberOfGuests
         } = {},
         session
       ) => {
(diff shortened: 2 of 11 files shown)


src/config/product.openshiftContainer.js
```diff
@@ -213,7 +214,7 @@ const config = {
               isInline
               component="a"
               variant="link"
-              href={`${helpers.UI_DEPLOY_PATH_LINK_PREFIX}/insights/inventory/${instanceId}/`}
+              href={`${helpers.UI_DEPLOY_PATH_LINK_PREFIX}/insights/inventory/${inventoryId}/`}
```
cdcabrera, line 217: additional updates, maybe?

starting around line `197` and going down to line `209` there is a conditional and ternary that falls back to displaying the `instanceId`... all of it with the intent of being a fallback the user could reference. If `instanceId` is no longer something a user can use to navigate then it should also be updated.

### from
```
{
  [INVENTORY_TYPES.DISPLAY_NAME]: displayName,
  [INVENTORY_TYPES.INVENTORY_ID]: inventoryId,
  [INVENTORY_TYPES.INSTANCE_ID]: instanceId,
  [INVENTORY_TYPES.NUMBER_OF_GUESTS]: numberOfGuests
} = {},
...
if (!instanceId) {
  return displayName;
}

let updatedDisplayName = displayName || instanceId;
```

### to (up to 3 additional updates)
```
{
  [INVENTORY_TYPES.DISPLAY_NAME]: displayName,
  [INVENTORY_TYPES.INVENTORY_ID]: inventoryId,
  [INVENTORY_TYPES.NUMBER_OF_GUESTS]: numberOfGuests
} = {},
...
if (!inventoryId) {
  return displayName;
}

let updatedDisplayName = displayName || inventoryId;
```

This change should also be applied to the other configs that also leverage that fallback.

src/config/product.openshiftMetrics.js
```diff
@@ -243,7 +244,7 @@ const config = {
               isInline
               component="a"
               variant="link"
-              href={`${helpers.UI_DEPLOY_PATH_LINK_PREFIX}/openshift/details/${instanceId}`}
+              href={`${helpers.UI_DEPLOY_PATH_LINK_PREFIX}/openshift/details/${inventoryId}`}
```
cdcabrera, line 247: question:
The SWATCH issue specifically calls out RHEL. You confirmed `OpenShift Dedicated` and `RHACS` in your PR/MR notes, but not `OpenShift Metrics`... is this correct and the notes just skipped it?

cdcabrera (changes requested):
Almost there... a few more updates per config

src/config/product.openshiftMetrics.js
Sgitario: I've just confirmed that the OpenShift Metrics should also use the inventory ID. Description updated.

src/config/product.openshiftContainer.js
Sgitario: To sum up, stop displaying the instance ID (which changed) and print the inventory ID, so users can use it if needed +100

Sgitario:
@cdcabrera PR updated

src/config/product.openshiftContainer.js
cdcabrera: You got it!

PR #1335, opened 2024-06-11
feat(config): sw-2402 activate rhel for x86 payg addon

cdcabrera:
## What's included

- feat(config): sw-2402 activate rhel for x86 payg addon

### Notes
- activates the RHEL for x86 payg addon as a variant of RHEL ELS x86 PAYG identifier, see screenshot for variant layout
- adjusts the copy for the original RHEL for x86 payg implementation to be more reflective of what it is
   - this copy is subject to confirmation by documentation

### Local run check
1. `$ npm start`
1. navigate to the rhel view
   - Confirm the original RHEL x86 ELS PAYG has had it's copy updated in the variant drop down list AND still fires the original product identifier
   - Confirm the new RHEL x86 ELS PAYG ADDON now HAS the original label used by RHEL x86 ELS PAYG but instead fires the new product identifier

## Example
_Note: numeric data may, or may not be, accurate during local development_

### Updated
Screenshot 2024-06-14 at 10 08 07 AM
Screenshot 2024-06-14 at 10 07 54 AM

Screenshot 2024-06-11 at 7 36 34 PM
Screenshot 2024-06-13 at 12-49-37 Standalone Curiosity

### Previous
Screenshot 2024-06-11 at 7 08 32 PM

## Updates issue/story

sw-2402


jlprevatt:
As I mentioned in our Slack DM 06/13/2024, I don't think that dropping the 3 characters from the product name (plus adding one back) from "Third" to "3rd" is gaining you anything significant, the name is already long. And it results in a mismatch in the product name from other places (in the HCC or other) that the customer might see it, especially if the name is read directly from the SKU metadata in any of the places it might appear. So I vote for "Third."

PR #1361, opened 2024-07-16
Schedule Jenkins pipelines in rhel8 nodes

rporres:
"insights" nodes have been decomissioned since they were based on CentOS 7

## What's included

Fix to the Jenkins file so that pipelines are scheduled in existing nodes.

## Example

...


maknop (approved):

cdcabrera:
@maknop and @rporres in the future please adhere to the commit checks...

PR #1416, opened 2024-09-17
3 of 4, refactor: sw-2479 replace blur with 5xx error display

cdcabrera:
## What's included

- refactor: sw-2479 replace blur with 5xx error display

### Notes
- Replaces the blur for a 5xx level unavailable display across the primary components, graphCard, inventoryCard, and inventory guests
- A generic message is initially displayed to align with the generic error displays in other areas of consoledot
   - Each error card contains a `gear` icon in the upper right that can be clicked to display the immeadiate error information associated with its related API calls. 
      - selecting the `textarea` and typing the keyword brings up the debugger log download link. This is the entire sanitized gui state log we use for external users, and is the same exact log across all links (no need to download it multiple times unless a new error happens after you've already downloaded it). See animation below
- The gui version information is now displayed in the lower right corner of the display. This links back to the gui public changelog. We used the lower-right due to avoid consoledot notification/icon links, however the banner that happens from Pendo notifications may still cover this area
- ~This logic assumes that if one rhsm API call is returning a 5xx level error they all are~
   - ~Based on individual product configuration we currently don't track how many rhsm API calls each product attempts.~ ~We could attempt to start tracking that information but it would require re-architecting from where the 5xx error display happens~

1. `$ npm test`

### Local run check
1. `$ npm start`
1. update the `src/services/rhsmServices` file mocks with a `500` forced status
   - confirm that the product displays similar to the below example

### Proxy run check
1. `$ npm run start:proxy`
1. break any rhsm API response with a 500 level error
   - confirm that the product displays similar to the below example messaging

### Check the build
1. `$ npm run build`

## Example

Oct-02-2024 12-06-18

Screenshot 2024-10-02 at 12 04 19 PM
Screenshot 2024-10-02 at 12 04 04 PM
Screenshot 2024-10-02 at 12 04 29 PM
Screenshot 2024-10-02 at 12 04 42 PM
Screenshot 2024-10-02 at 1 07 30 PM
Screenshot 2024-10-02 at 12 09 06 PM

## Updates issue/story

sw-2479


cdcabrera:
Looks like it's possible for the different facets of the API to return different 5xx level errors. Have to reevaluate this blanket fix with a better solution. Automatically tracking the number of calls per product display then displaying the blanket page would probably work better

jlprevatt:
Adding a comment about the question/nitpick you missed in Slack:

Because the word "inventory" is not used on the subwatch screen to describe the tables, and we have other things that are actually called inventory (HBI/system inventory, subscriptions inventory), is the message for the tables correct if you use the word "inventory?"

IOW, is the error where the subwatch table data does not display caused **ONLY** by an error in those inventory services, or can it be caused by other problems? I don't think this is critical, but wanted to raise the question.

Another alternative that would work for either table is to say "Table display is ..."

PR #1526, opened 2025-02-18
chore: skip dependabot commit linting

kflahert:
## What's included

...
Skip linting for commit message length on dependabot PRs

## How to test

Skip commit message length for dependabot commit messages.

## Example

...


cdcabrera:
So it would be better to adjust dependabot to fit the script @kflahert ... if you're referencing the "react-18next" commit. The "group" naming combined with how dependabot groups the patch and minor commits just needs to be broken out similar to the dev deps

PR #1542, opened 2025-03-04
fix(inventoryGuests): sw-3216 infinite loader

cdcabrera:
## What's included

- fix(inventoryGuests): sw-3216 infinite loader

## Example

...

## Updates issue/story

sw-3216


vbusch:
@cdcabrera this looks like it is ready for review.   Should it be marked as 'ready for review', so it is no longer in 'Draft'?

cdcabrera:
@vbusch not yet, we were concerned the solution wasn't viable

mmusil (approved):
LGTM, automation PR is also ready.

PR #1581, opened 2025-04-23
fix(toolbarFieldBillingAccount): sw-3513 remove locale ref

cdcabrera:
## What's included

- fix(toolbarFieldBillingAccount): sw-3513 remove locale ref

### Notes
- In scenarios where account identifiers contain underscores they can alter how the translation wrapper determines context. The fix is to avoid using the translation wrapping entirely.

### Local run check
1. update the NPM packages with `$ npm install`
1. `$ npm start`
1. navigate to a product display, such as OpenShift AI
   - confirm the billing provider and billing account fields are displaying as intended

### Proxy run check
1. update the NPM packages with `$ npm install`
1. make sure you're on network, then
1. `$ npm run start:proxy`
1. navigate to a product display, such as OpenShift AI
   - confirm the billing provider and billing account fields are displaying as intended

## Example

...

## Updates issue/story

sw-3513
related to sw-95 #1577

PR #1583, opened 2025-04-23
fix(toolbar): sw-3513 conditional locale display

cdcabrera:
## What's included

- fix(toolbar): sw-3513 conditional locale display

### Notes
- Applies a conditional check for "dynamic" toolbar chip/label values. The same issue that was happening in the original sw-3513 was also happening at the chip/label level, see examples below

### Local run check
1. `$ npm start`
1. navigate towards a product that uses the billing provider filter
   - confirm that after selecting an account id that contains multiple underscores the related toolbar chip/label also displays correctly

### Proxy run check
1. `$ npm run start:proxy`
1. navigate towards a product that uses the billing provider filter
   - confirm that after selecting an account id that contains multiple underscores the related toolbar chip/label also displays correctly

## Example

After
Screenshot 2025-04-23 at 2 00 43 AM

Before
Screenshot 2025-04-23 at 1 59 00 AM

## Updates issue/story

sw-3513
related sw-95 #1581

### DIFF
--- a/src/components/README.md
+++ b/src/components/README.md
@@ -5597,14 +5597,16 @@ Set selected options for chip display.
 <table>
   <thead>
     <tr>
-      <th>Param</th><th>Type</th>
+      <th>Param</th><th>Type</th><th>Default</th>
     </tr>
   </thead>
   <tbody>
 <tr>
-    <td>params</td><td><code>object</code></td>
+    <td>params</td><td><code>object</code></td><td></td>
     </tr><tr>
-    <td>params.value</td><td><code>string</code> | <code>Array.&lt;string&gt;</code></td>
+    <td>params.value</td><td><code>string</code> | <code>Array.&lt;string&gt;</code></td><td></td>
+    </tr><tr>
+    <td>[params.isLocale]</td><td><code>boolean</code></td><td><code>true</code></td>
     </tr>  </tbody>
 </table>
 
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -103,9 +103,10 @@ const Toolbar = ({
    *
    * @param {object} params
    * @param {string|Array<string>} params.value
+   * @param {boolean} [params.isLocale=true]
    * @returns {Array}
    */
-  const setSelectedOptions = ({ value: filterName } = {}) => {
+  const setSelectedOptions = ({ value: filterName, isLocale = true } = {}) => {
     const filterValues =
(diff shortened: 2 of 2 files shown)


src/components/README.md
```diff
@@ -5597,14 +5597,16 @@ Set selected options for chip display.
 <table>
   <thead>
     <tr>
-      <th>Param</th><th>Type</th>
+      <th>Param</th><th>Type</th><th>Default</th>
```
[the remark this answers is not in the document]
cdcabrera: you're reading an auto generated file you garbage bot

src/components/toolbar/toolbar.js
```diff
@@ -115,7 +116,12 @@ const Toolbar = ({
 
     return filterValues
       .filter(({ value }) => typeof value === 'string')
-      .map(({ name, value }) => t('curiosity-toolbar.label', { context: [name, (value === '' && 'none') || value] }));
+      .map(
+        ({ name, value }) =>
+          (isLocale && t('curiosity-toolbar.label', { context: [name, (value === '' && 'none') || value] })) ||
+          (value === '' && 'none') ||
```
cdcabrera, line 122: optional: it comes across as ignoring the `isLocale` check, but it may be good to use a locale string for `none`... as it is currently, this would be one of the few display strings we hardcode

cdcabrera: we ended up having to allow optional level configuration instead of having a general condition check. there's a subtle change that's highlighted in the screenshot where the `aws` is displayed instead of the full locale string

src/components/toolbar/toolbar.js
```diff
@@ -86,7 +86,23 @@ const Toolbar = ({
...
+          if (typeof obj === 'string') {
+            return obj;
+          }
+          return obj.name;
+        })
+      );
+    }
+
     updatedValue.forEach(val => clearField(val));
   };
```
cdcabrera, line 107: future: refactor should include making some flavor of this "on clear" func dep injected for testing

src/components/toolbar/toolbar.js
```diff
@@ -102,20 +118,41 @@ const Toolbar = ({
...
+
+          const { isDisplayValueOnly, name } = filter;
+          return {
+            isDisplayValueOnly,
+            name,
+            value: toolbarFieldQueries[name]
+          };
+        })
+      );
+    }
```
cdcabrera, line 146: future: refactor should include making some flavor of this "setting selected options" dep injected for testing

PR #1587, opened 2025-04-26
feat(rhsmTransformers,bannerMessages): sw-3519 usage alert banner

cdcabrera:
## What's included

- feat(rhsmTransformers,bannerMessages): sw-3519 usage alert banner

### Notes
- The 3 primary scenarios we cover for displaying the banner are
   - The are multiple providers with 1 or more accounts
      - The base message is `Usage that has no matching subscription is being reported for {{provider}}) billing account {{account}}, and {{remaining}}. Learn more...`
   - There is a single provider with 1 or more accounts
      - The base message is `Usage that has no matching subscription is being reported for {{provider}}) billing account {{account}}, and {{remaining}}. Learn more...`
   - There is a single provider with a single account
      - The base message is `Usage that has no matching subscription is being reported for {{provider}}) billing account {{account}}. Learn more...`
- There is a fallback message for scenarios that fail to meet the checks in the form of
   - `Usage that has no matching subscription is being reported. Learn more...`

### Proxy run check
1. `$ npm run start:proxy`
1. navigate to a pay as you go product variant
   - make sure that variant has no related subscriptions for it's instances/usage (current accounts in staging seem to not have subscriptions, just usage =) )
   - the banner should display based on the above noted scenarios

## Example

Screenshot 2025-04-26 at 11 39 01 AM

## Updates issue/story

related sw-95 #1586 
sw-3519


src/components/bannerMessages/bannerMessagesContext.js
```diff
@@ -91,30 +91,34 @@ const useSetBannerMessages = ({
...
+                id: value,
+                title: value
+              };
+            }
+
+            return undefined;
+          })
+          .filter(value => value !== undefined);
+
+        const updatedBannerMessages = bannerMessages?.filter(({ id: existingId }) =>
```
cdcabrera, line 114: required: confirm this filter with some... think it's supposed to be 
```
const updatedBannerMessages = bannerMessages?.filter(({ id: existingId }) =>
  !updatedMessages.some(({ id: newId }) => newId === existingId)
 );
```

cdcabrera: changed, unit test exposed this

src/services/rhsm/__tests__/__snapshots__/rhsmHelpers.test.js.snap
```diff
@@ -0,0 +1,133 @@
...
+    ],
+    "firstProvider": "lorem",
+    "firstProviderAccount": "123",
+    "firstProviderNumberAccounts": 3,
+    "hasUniqueAccounts": true,
+    "hasUniqueProviders": false,
+    "numberProviders": 1,
+    "providers": [],
+  },
+}
```
cdcabrera, line 95: required: this isn't technically the scenario we intended to return... it should be a unique provider, note the `hasUniqueProviders` flag

confirm the scenario in test, add an additional scenario where the provider is unique but still falls into the test description

cdcabrera: added to scenario variations, results and descriptions align to what is intended

PR #1618, opened 2025-05-29
fix(config): sw-3618 use vCPUs for rhacm

lindseyburnett:
## What's included

- fix(config): sw-3618 use vCPUs for rhacm

### Notes
- moves from cores to the specific vcpus metric

1. `$ npm test`

### Proxy run check
1. `$ npm run start:proxy`
1. confirm vcpus are coming through for the RHACM variant

## Example

...

## Updates issue/story

sw-3618


src/config/product.rhacm.js
```diff
@@ -102,7 +102,7 @@ const config = {
     {
       filters: [
         {
-          metric: RHSM_API_PATH_METRIC_TYPES.CORES,
+          metric: RHSM_API_PATH_METRIC_TYPES.VCPUS,
```
cdcabrera, line 105: required: looks like we missed the locale strings, need to add them for the new metric

cdcabrera (changes requested):

cdcabrera (approved):

PR #1626, opened 2025-06-06
1 of 2, refactor(deps): sw-3639 pf6 updates

cdcabrera:
## What's included

- refactor(deps): sw-3639 pf6 updates
- fix: sw-3639 notifications autoDismiss only
- ~refactor(components): sw-3639 pf6 component adjustments~ (subsequent pr)

### Notes
- bundle related styling and components affected by pf6
- see below comment on why we're disabling toast notification "dismiss Xs"
- ~nice-to-have component adjustments that affect the display in minimally noticeable ways~

## Example

Screenshot 2025-06-06 at 14-22-07 Standalone Curiosity
Screenshot 2025-06-06 at 14-22-27 Standalone Curiosity
Screenshot 2025-06-06 at 14-23-23 Standalone Curiosity
Screenshot 2025-06-05 at 2 51 04 PM

## Updates issue/story

sw-3639

PR #1629, opened 2025-06-13
perf(components): sw-3639 pf6 component adjustments

cdcabrera:
## What's included

- perf(components): sw-3639 pf6 component adjustments
   * chart, css for smoother loading
   * select, css for smoother loading
   * graphCardChart, streamline ternary for loading
   * toolbarFields, isInline prop, responsive display related

### Notes
- updates related to enhancing component loading
   - select was displaying the "disabled" select, we've added a css fade-fast to help reduce the jarring "disabled -> loaded" display
   - chart redraw has always been funky, this just adds a css fade to "mask" the svg load
- toolbarFields, expose the inline block prop for "future us" and responsive layout adjustments

## Example

...

## Updates issue/story

sw-3639
related #1626


cdcabrera:
Looks like we have a modal position issue for the subs mismatch banner at "small" to "medium" screen sizes
Screenshot 2025-06-16 at 2 07 01 PM

cdcabrera:
The modal issue looks like it's related to
- the element we're positioned off of
- styling outside our control combined with the way the chroming responsive design behaves in browser chrome vs firefox

As a first step we're removing the deprecated modal in favor of the full pf6 version, maybe it'll kick things into gear, but we won't know for sure until the work merges. After that we'll attempt to append the modal to a different element

PR #1633, opened 2025-06-17
ci: sw-3595 rm swatch-producer-red-hat-marketplace

vbusch:
## What's included
swatch redhat marketplace is no longer deployed.  Remove it from the ci.  This will enable the ephemeral deploy to work.
swatch-3595
...

## Example

...

## Updates issue/story

sw-3595


lindseyburnett (approved):

cdcabrera:
@vbusch "swatch" is a keyword that triggers linking over to the issue board. recommend you update the commit message. probably should mention "ephemeral" in there since it's still a little vague as to what we're doing here... there's some older log messages around this

PR #1646, opened 2025-07-03
ci: Define konflux pipelines again

lindseyburnett:
Used generated pipeline from PR #1645 

Then added the "stable" branch


vbusch (approved):

cdcabrera:
@lindseyburnett  for clarity I've only come across a handful of times this has been issue. And it always seems to be related to some aspect of manually getting Konflux to update its process, like we've forced/pushed an update. I'd like to know more before I consider this

cdcabrera (approved):

PR #1863, opened 2026-05-20
SWATCH-4975: Create Konflux Pipeline for hermetic frontend build

Sgitario:
## What's included
Starting May 13, the frontend will need to be build with the Nexus registry in order to be released.  Or a prodsec-approvaed policy exception will be required.

There is already a [Dockerfile.hermetic](https://github.com/RedHatInsights/insights-frontend-builder-common/blob/c1204294e120df77acf22edc24828598da4a1702/Dockerfile.hermetic) in the build-tools submodule.  And a [README-hermetic-build.md](https://github.com/RedHatInsights/insights-frontend-builder-common/blob/master/README-hermetic-build.md).  We need to create a pipeline that uses the Dockerfile.hermetic instead of the Dockerfile.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -45,4 +45,4 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.COV_TOKEN }}
       - name: Confirm integration
         if: ${{ success() }}
-        run: npm run build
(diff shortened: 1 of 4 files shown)


Sgitario:
In draft, until the ephemeral deployment gets fixed to support hermetic builds.

PR #1873, opened 2026-06-04
ci: configure bonfire iqe pipeline

Sgitario:
## What's included
Move the konflux pipelines into the curiosity repository, plus manage the summary comments from konflux.


mmusil (approved):
lgtm

PR #1881, opened 2026-06-23
fix(ci): use identity auth for Bonfire ephemeral IQE

Sgitario:
## Summary

- Pass `IQE_ENV_VARS=DYNACONF_HTTP__DEFAULT_AUTH_TYPE=identity` into the Bonfire `run-iqe-cji` task for `curiosity-frontend-bonfire-tekton`.
- Fixes `KeyError: 'refresh_token'` in ephemeral UI tests: `curiosity_user_app` calls `rhsm_subscriptions` REST APIs during fixture setup, but `ENV_FOR_DYNACONF=ephemeral` inherits `jwt-auth` from iqe-core while clowder users only have username/password (no refresh token).
- rhsm-subscriptions GitLab CI avoids this by running marker `ephemeral` with `IQE_ENV=clowder_smoke` (identity auth in YAML). Bonfire Curiosity must use `ephemeral` for Playwright outside the cluster; this override aligns API auth without changing `iqe-rhsm-subscriptions-plugin`.

## Test plan

- [ ] Konflux `curiosity-frontend-bonfire-tekton` IntegrationTestScenario passes on this PR
- [ ] IQE pod log shows `Setting auth_type to identity` (not `jwt-auth` / `refresh_token`) before `put_opt_in_config`
- [ ] Ephemeral UI tests (`@pytest.mark.ephemeral`) complete past fixture setup

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.tekton/bonfire-integration-tests-pipeline.yaml
+++ b/.tekton/bonfire-integration-tests-pipeline.yaml
@@ -118,6 +118,13 @@ spec:
       type: string
       description: Value to set for ENV_FOR_DYNACONF
       default: "ephemeral"
+    - name: IQE_ENV_VARS
+      type: string
+      description: >-
+        Comma-separated extra env vars for the IQE pod (see bonfire-tekton run-iqe-cji).
+        Ephemeral Bonfire runs outside the cluster; use identity auth for rhsm_subscriptions
+        API setup (clowder_smoke sets this in YAML; ephemeral does not).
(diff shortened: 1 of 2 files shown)


Sgitario:
Note that this is not a solution, but I only want to quickly test this change. Right solution should be done in the iqe rhsm subscriptions plugin.

PR #1888, opened 2026-06-29
SWATCH-5091:fix export notification logic

Mantzounis:
## What's included
This PR fixes export notification scoping to current product page only.
Part of the acceptance criteria is to SWATCH-4427 IQE test `test_export_subscription_data_json_manual_download` passes after the fix (or is updated if behavior intentionally changes). however due to the issues in ephemeral this will be verified in stage after merging. Also worth noting is that a video of the issue is attached to the [ticket](https://redhat.atlassian.net/browse/SWATCH-5091).
...

1. `$ npm test`

### Local run check
2. `$ npm start`
3. Temporarily edit `src/services/platform/platformServices.js`:
     - Remove `@apiMock {RandomSuccess}` from the `GET /api/export/v1/exports` JSDoc
     - In the first `@apiSuccessExample`, remove the `swatch-RHEL for x86` object,
       leaving only `swatch-rhel-for-x86-els-payg` with `"status": "complete"`
4. Navigate to the **RHEL for x86** product page (parent)
5. Verify notification doesn't appear
6.  Navigate to the ELS On-Demand for Third Party Linux Migration variant page (rhel-for-x86-els-payg)
7. Verify: notification appears
### Proxy run check
2. make sure you're on network, then
3. `$ npm run start:proxy`
4. Navigate to a product page, trigger an export on a different product and confirm notification only appears on the product page where export was created

## Example

...

## Updates issue/story
updates SWATCH-5091

### DIFF
--- a/src/components/README.md
+++ b/src/components/README.md
@@ -6635,6 +6635,8 @@ Apply an existing exports hook for user abandoned reports. Allow bulk polling st
     </tr><tr>
     <td>options.useNotifications</td><td><code>NotificationsContext.useNotifications</code></td>
     </tr><tr>
+    <td>options.useProduct</td><td><code>useProduct</code></td>
+    </tr><tr>
     <td>options.useSelectorsResponse</td><td><code>storeHooks.reactRedux.useSelectorsResponse</code></td>
     </tr>  </tbody>
 </table>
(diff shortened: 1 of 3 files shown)


vbusch (changes requested):
I have a different understanding of the expected results.
After requesting the export, the notification should always appear, regardless of which product you are current viewing.

PR #1903, opened 2026-07-03
test(toolbar): add export unit tests.

Mantzounis:
## What's included
After the investigation of the [card](https://redhat.atlassian.net/browse/SWATCH-4427) we have decided to remove the iqe test mentioned on the card and this PR aims to cover the test path of export as best as possible with unit tests. After this merges the integration test will be removed in this [MR](https://gitlab.cee.redhat.com/insights-qe/iqe-curiosity-plugin/-/merge_requests/635). 
...

### Interactive unit test check
1. `$ npm run test:dev`

## Example

...

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for export-related toolbar behavior, including cache handling, notification display/removal, and mount/unmount behavior.
  * Added more detailed checks for pending vs. completed export states and how existing notifications affect new alerts.
  * Updated confirmation flow coverage to verify the expected export lists are passed through correctly.

### DIFF
--- a/src/components/toolbar/__tests__/__snapshots__/toolbarFieldExportContext.test.js.snap
+++ b/src/components/toolbar/__tests__/__snapshots__/toolbarFieldExportContext.test.js.snap
@@ -461,48 +461,6 @@ exports[`ToolbarFieldExport Component should allow and export, and expose pollin
 }
 `;
 
-exports[`ToolbarFieldExport Component should allow export service calls on existing exports: existingExports 1`] = `
-[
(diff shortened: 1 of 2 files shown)


vbusch (approved):
It follows the existing pattern and provides good coverage, with clear descriptions.

It adds 3 new snapshot comparison tests, which is the exiting pattern.  But those three could have avoided the use of snapshots.  Which would make the asserts easier to ready and less brittle.

PR #1965, opened 2026-08-21
fix(app): resolve AppEntry TDZ from Kessel circular init

diegomaranhao:
## What's included
fix(app): resolve AppEntry TDZ from Kessel circular init
Hoist AppEntry as a function declaration and lazy-load App so Chrome can
resolve the federated export before Kessel/auth evaluate. Install
@project-kessel/react-kessel-access-check as a real dependency instead of
an optional peer.

## Example

...


vbusch (commented):
@diegomaranhao can you give more context as to which environment needs this change. 
Adding React.lazy adds more complexity.

diegomaranhao (commented):
@vbusch I was having a circular issue before with the fix of https://github.com/RedHatInsights/curiosity-frontend/pull/1972 to run it on stage proxy `npm run start:proxy` , but the last changes seem to have fixed the issue. I'm closing this PR. Thanks

PR #1972, opened 2026-08-24
SWATCH-5457: fix(authentication): swatch-5457 overlay inventory for v2 orgs

diegomaranhao:
## What's included
 Two related authentication fixes for Kessel/v2 org environments.
     
  authentication.js — overlay inventory in session authorized

  Instance table instance links read session.authorized.inventory to decide whether to render. In v2
  orgs (Kessel enabled or not), the RBAC v1 inventory permission is always empty, so those links were
  silently hidden even for authorized users. When isAuthorized is true, inventory: true (and appName: 
  true) are now overlaid onto the context value. The constructed object is wrapped in useMemo to avoid
  recreating the context reference on every render. A helpers.browserExpose({ authDebug }) block is
  included for in-browser debugging.

  useHasRelation.js — add enabled to useEffect deps

  The Kessel feature flag (swatch.common-security.use-kessel-rbac) is read from Chrome after the first
  paint. With an empty [] dep array the check ran once on mount — before the flag had resolved — so
  enabled was always false and checkSelf was never called. Adding enabled to the dep array lets the
  effect re-run when the flag arrives.

## How to test
To make `$npm run start:proxy` work, you need the following fix https://github.com/RedHatInsights/curiosity-frontend/pull/1965

Coverage and basic unit test check

  1. $ npm install
  2. $ npm test

  Local run check (mock API, dev mode — Kessel path skipped, legacy RBAC exercised)

  1. $ npm install
  2. $ npm start
  3. Navigate to any product page and confirm instance table links are visible

  Proxy run check (exercises Kessel path end-to-end)

  1. $ npm install
  2. $ npm run start:proxy
  3. In a v2 org with Kessel flag on, navigate to a product page
  4. Open the browser console → window.__curiosity.authDebug should show isAuthorized: true and
  authorized.inventory: true
  5. Confirm instance table instance links render correctly

## Example
N/A — auth/session context fix; no visual diff on the component itself.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authentication state handling to consistently include application and inventory access when authorized.
  - Authentication details are now available throughout the app and browser integrations.
  - Relation checks now re-run when authentication becomes enabled, including after feature flags load.

- **Tests**
  - Added coverage for authorized access across legacy and updated authentication paths.
  - Added coverage for relation checks triggered when authentication is enabled.

### DIFF
--- a/src/components/authentication/__tests__/__snapshots__/authentication.test.js.snap
+++ b/src/components/authentication/__tests__/__snapshots__/authentication.test.js.snap
@@ -5,6 +5,7 @@ exports[`Authentication Component should allow being disabled: disabled 1`] = `
   value={
     {
       "authorized": {
+        "inventory": true,
         "subscriptions": true,
       },
       "errorCodes": [],
@@ -43,6 +44,7 @@ exports[`Authentication Component should render a component authorized: authoriz
   value={
     {
       "authorized": {
+        "inventory": true,
         "subscriptions": true,
       },
       "errorCodes": [],
@@ -98,7 +100,10 @@ exports[`Authentication Component should render authorized via kessel when flag 
 <Context.Provider
   value={
     {
-      "authorized": {},
+      "authorized": {
+        "inventory": true,
+        "subscriptions": true,
+      },
       "errorCodes": [],
       "errorStatus": undefined,
     }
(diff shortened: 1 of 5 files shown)


vbusch (approved):
LGTM

Tested in ephemeral with `swatch.common-security.use-kessel-rbac` on and off

